PR #11, opened 2018-03-22
Feature/brbdcf 1043 hosts pool from file

laurentganne:
# Pull Request description

## Description of the change

Added a command to apply a Hosts Pool configuration provided in a YAML or JSON file :
```bash
yorc hostspool apply [--auto-approve] <filename>
```
And a command to export the Hosts Pool configuration as a YAML or JSON representation, to the standard output or a file:
```bash
yorc hostspool export [--output yaml|json] [--file <filename>]
```

The apply command  accepts YAML or JSON files. The following properties are supported :

  * ``hosts``: List of hosts configuration. A host configuration supports the following properties,
     - ``name``: mandatory string identifying the host, no other host entry can have the same name value in the file
     - ``connection``: Connection configuration,
        + ``host``: Hostname or ip address used to connect to the host (defaults to the ``name`` described above)
        + ``user``: name of the user used to connect to the host (default "root")
        + ``password``: either a password or a private key should be provided
        + ``private_key``: Path to a private key file (or private key file content), either a password or a private key should be provided
        + ``port``: Port used to connect to the host (default 22)
     - ``labels``: key/value pairs

Example of a YAML Hosts Pool configuration file :

```YAML

    hosts:
    - name: host1
      connection:
        host: host1.example.com
        user: test
        private_key: /path/to/secrets/id_rsa
        port: 22
      labels:
        environment: dev
        testlabel: hello
        host.cpu_frequency: 3 GHz
        host.disk_size: 50 GB
        host.mem_size: 4GB
        host.num_cpus: "4"
        os.architecture: x86_64
        os.distribution: ubuntu
        os.type: linux
        os.version: "17.1"
    - name: host2
      connection:
        host: host2.example.com
        user: test
        password: test
```

The apply will compare and display the differences between the current Hosts Pool configuration and the configuration specified in the file. A user confirmation will be asked before proceeding, except if --auto-approve option is specified.

### How I did it

In prov/hostspool/hostspool_mgr.go, added a new function Apply(pool []Host, checkpoint *uint64)
This function will compare the current Hosts Pool configuration with the Hosts Pool in argument and apply changes. The checkpoint in argument, if not nil, will be used to verify the changes are applied to the expected Hosts Pool configuration. The checkpoint value for the current Hosts Pool configuration is returned by the function List() implemented in the same file.
Internally, this checkpoint corresponds to a consul ModifyIndex. 
As a consequence the lock key kvLocckKey initially created under the hosts_pool subtree had to be moved outside of this subtree, or the command apply by just locking this lock was changing the ModifyIndex, making the checkpoint in argument invalid.

Factorized the code already existing for CRUD operations on one host, to have common code computing necessary  operations that the new Apply function will apply in a single transaction.

### How to verify it

Preliminary step: environment setup

To test a Hosts Pool in a development environment, you can use a dockerized ssh server.
First as a yorc user, generate private/public keys in a <path to secrets>/ directory, check the private key is read-only for your yorc user,
then run the following commands to have 3 test hosts host1, host2, host3 that can be added to a Hosts Pool :
(--privileged option below is needed because the container has also docker-in-docker capability, although it is not used here)
```bash
docker run --privileged -p 8701:22 \
 -e "AUTH_KEY=$(cat <path to secrets>/id_rsa.pub)" \
 --rm -d --name host1 --hostname host1 laurentg/testhost

docker run --privileged -p 8702:22 \
 -e "AUTH_KEY=$(cat <path to secrets>/id_rsa.pub)" \
 --rm -d --name host2 --hostname host2 laurentg/testhost

docker run --privileged -p 8703:22 \
 -e "AUTH_KEY=$(cat <path to secrets>/id_rsa.pub)" \
 --rm -d --name host3 --hostname host3 laurentg/testhost
```

Verify that as yorc user, you can connect to each host using the private key you generated :
```bash
ssh -i <path to secrets>/id_rsa test@myhost -p 8701
ssh -i <path to secrets>/id_rsa test@myhost -p 8702
ssh -i <path to secrets>/id_rsa test@myhost-p 8703
```

Hosts Pool apply tests

You can get examples of yaml/json files already created and update connection settings to use your host:

Configuration with 1 host : https://github.com/laurentganne/ystia-devenv/blob/master/testlab/pool1host.yaml
Configuration with an additional host and label changes for the existing host : https://github.com/laurentganne/ystia-devenv/blob/master/testlab/pool2hosts.yaml
Same configuration with label diffs, in json : https://github.com/laurentganne/ystia-devenv/blob/master/testlab/pool2hosts.json
Configuration with an additional host : https://github.com/laurentganne/ystia-devenv/blob/master/testlab/pool3hosts.yaml

First apply pool1host.yaml :

```bash
yorc hp apply pool1host.yaml
```

Check you get a confirmation message showing host1 will be added.
Enter y

Check you have 1 host now in the pool, running :

```bash
yorc hp list
```

Then apply pool2host.yaml

```bash
yorc hp apply pool2host.yaml
```

Check the confirmation message shows host2 will be added, and 2 host1 labels will be updated and 1 host1 label will be removed
Enter y

Then apply pool2host.json

```bash
yorc hp apply pool2host.json
```

Check there will be label changes on host1 and host2

Then apply pool3host.yaml but don't answer to the confirmation question.
Just stay in this interactive step.

```bash
yorc hp apply pool3host.yaml
```

Instead of accepting the change, launch another window, and form this new window, launch a command to used the first configuration with 1 host :

```bash
yorc hp apply pool1host.yaml
```

Check you are warned that host2 will be removed. Enter y.

Go back to the first window where you attempted to add a third host to the pool that had 2 hosts at that time.
Enter y
Check you get an error message that detected a Hosts Pool checkpoint issue, as the Hosts Pool configuration has changed since, and the operation is refused.

Try again the command :

```bash
yorc hp apply pool3host.yaml
```

Check the confirmation message correctly states that host2 and host3 will be added.
Enter n (or nothing) to refuse the change.

Stop the host2 running this command :

```bash
docker stop host2
```

Try to add host2 to the pool :

```bash
yorc hp apply pool2host.yaml
```
Enter y

Check you get a message that the apply was done successfully but that a connection failure was detected for host2.

Start again host2 :

```bash
docker run --privileged -p 8702:22 \
 -e "AUTH_KEY=$(cat <path to secrets>/id_rsa.pub)" \
 --rm -d --name host2 --hostname host2 laurentg/testhost
```

Verify that as yorc user, you can connect to host2: 
```bash
ssh -i <path to secrets>/id_rsa test@myhost -p 8702
```

Apply the json hosts pool configuration that just updated labels :

```bash
yorc hp apply pool2hosts.json
```

Run yorc hp list and check host2 is back online with no connection error

Export checks :

Run this command to see the Hosts Pool configuration exported in YAML representation :

```bash
yorc hp export
```

Run this command to have a json output :

```bash
yorc hp export -o json
```

Run this command to save the Hosts Pool configuration in a yaml file :

```bash
yorc hp export -f tmp.yaml
```

### DIFF
--- a/prov/hostspool/hostspool_mgr_test.go
+++ b/prov/hostspool/hostspool_mgr_test.go
@@ -521,15 +521,12 @@ func testConsulManagerConcurrency(t *testing.T, cc *api.Client) {
 	assert.Error(t, err, "Expecting concurrency lock for releaseWait()")
 }
 
-func testConsulManagerApply(t *testing.T, cc *api.Client) {
-	cleanupHostsPool(t, cc)
-	cm := &consulManager{cc, mockSSHClientFactory}
+func createHosts(hostsNumber int) []Host {
 
-	var hostpool []Host
-
-	for i := 0; i < 3; i++ {
+	var hostpool = make([]Host, hostsNumber)
+	for i := 0; i < hostsNumber; i++ {
 		suffix := strconv.Itoa(i)
-		hostpool = append(hostpool, Host{
+		hostpool[i] = Host{
 			Name: "host" + suffix,
 			Connection: Connection{
 				User:     "testuser" + suffix,
@@ -542,8 +539,16 @@ func testConsulManagerApply(t *testing.T, cc *api.Client) {
 				"label2": "value2" + suffix,
 				"label3": "value3" + suffix,
 			},
-		})
+		}
 	}
+
+	return hostpool
+}
+func testConsulManagerApply(t *testing.T, cc *api.Client) {
+	cleanupHostsPool(t, cc)
+	cm := &consulManager{cc, mockSSHClientFactory}
+
+	var hostpool = createHosts(3)
 
 	// Apply this definition
 	var checkpoint uint64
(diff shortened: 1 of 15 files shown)


prov/hostspool/hostspool_mgr_test.go
```diff
@@ -502,3 +520,252 @@ func testConsulManagerConcurrency(t *testing.T, cc *api.Client) {
 	err = cm.releaseWait("concurrent_host1", 500*time.Millisecond)
 	assert.Error(t, err, "Expecting concurrency lock for releaseWait()")
 }
+
+func testConsulManagerApply(t *testing.T, cc *api.Client) {
```
stebenoist: Maybe this test can be split in order to improve the test excution time ?

laurentganne: ok ! Will split the test into several tests organized by topic

loicalbertin:
Sounds almost perfect to me just two points that may be improved in my sense:

- Error message for check point issues sounds good to me for the REST API point of view but may be improve in the CLI to help the user understand what's going on and how to fix it. Maybe something like "Hosts Pool changed since the above diff. You should reapply your configuration to see actual diff." would be helpful

- Then for the case where the connection to the host failed, from my understanding currently the only way to make the host available again is to change it (using labels or connection fields). But I run into an issue were the connection issue was not related to the hostspool config by itself the key file was missing on the host. To fix it I just  make it available at the expected location. And finally I was expecting to not change the hostpool configuration as it is correct actually, but just retry the connection. WDYT?

laurentganne:
Ok ! Will change the CLI error message for the checkpoint issue.

Regarding the connection failure, yes in this implementation if the connection failure is not due to a Hosts Pool config issue, then you won't need to run a 'hostspool apply' command, and so the Host status will still appear as down in the Hosts Pool.

We could decide that the 'hostspool apply' which today does nothing for Hosts with no config change, would take care of verifying connections for such hosts with no config change, like what is already done for hosts with config change.

Another proposal would be to not develop anything for this case here, but to have this case managed by the not yet implemented User Story on auto-healing policies and Compute liveness.
Would it be ok ?

loicalbertin:
> Another proposal would be to not develop anything for this case here, but to have this case managed by the not yet implemented User Story on auto-healing policies and Compute liveness.

I'm not sure that Compute liveness would help here as the host may be up&running but the configured credentials may be wrong. This is two different things IMHO.

loicalbertin:
We can also imagine a new REST endpoint that will allow to trigger a connection check on host(s)

laurentganne:
Ah yes, I was thinking in the case of Hosts Pool, the compute liveness would consist in connecting to the host, but indeed it could just be a ping or anything else.

For a simple change, but maybe not nice, we could have the 'hostspool apply' do the work of updating connections of all hosts in the pool even when there is no change,

Maybe a better solution would be to introduce a new command 'hostspool refresh' 
with this description:

hostspool refresh :
This command is used to reconcile the state Yorc knows about with the real-world infrastructure.
For example in case of a hosts connection failure happening at the time when a Host was added or updated in the Hosts Pool,
once the infrastructure issue is fixed,  this command can be run the update the connection status of this host in the Hosts Pool

loicalbertin:
Ok lets handle this case in a separate US as it is not only related to this one actually.
We can imagine that a password changed between the initial check and the first try to use a host. 

We will decide after having a bit more of feedback how to fix it properly.

PR #12, opened 2018-03-26
Ansible doesn't handle SSH connection with port different than 22

stebenoist:
## Description of the change
Set capability endpoint attribute port for host pool
Handle port different than 22 for Ansible for playbook execution

### How to verify it
Create a host pool specifying a port different than 22
Use a dockerized ssh server as laurentg/testhost image
docker run --privileged -p 8701:22 \
 -e "AUTH_KEY=$(cat <path to secrets>/id_rsa.pub)" \
 --rm -d --name host1 --hostname host1 laurentg/testhost

Deploy a topology with A4C with a simple component on this host pool and check Ansible use the defined port to SSH the host for installation

### DIFF
--- a/prov/ansible/execution.go
+++ b/prov/ansible/execution.go
@@ -103,6 +103,7 @@ type ansibleRunner interface {
 type executionCommon struct {
 	kv                       *api.KV
 	cfg                      config.Configuration
+	ctx                      context.Context
 	deploymentID             string
 	taskID                   string
 	NodeName                 string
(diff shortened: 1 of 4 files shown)


prov/ansible/execution.go
```diff
@@ -330,6 +332,17 @@ func (e *executionCommon) setEndpointCredentials(kv *api.KV, host, instanceID, c
 		if found && privateKey != "" {
 			conn.privateKey = config.DefaultConfigTemplateResolver.ResolveValueWithTemplates("host.privateKey", privateKey).(string)
 		}
+
+		found, port, err := deployments.GetInstanceCapabilityAttribute(e.kv, e.deploymentID, host, instanceID, "endpoint", "port")
+		if err != nil {
+			log.Debugf("[Warning] failed to get port attribute from capability endpoint with error: %+v", err)
```
loicalbertin: I think this should also be logged in deployment logs

loicalbertin: Same for the other debug log bellow

stebenoist: Yes, maybe we should return other errors around user, password and private key ?

prov/ansible/execution.go
```diff
@@ -303,44 +305,46 @@ func (e *executionCommon) resolveArtifacts() error {
...
 		found, port, err := deployments.GetInstanceCapabilityAttribute(e.kv, e.deploymentID, host, instanceID, "endpoint", "port")
 		if err != nil {
-			log.Debugf("[Warning] failed to get port attribute from capability endpoint with error: %+v", err)
+			return err
 		}
 		if found && port != "" {
 			conn.port, err = strconv.Atoi(port)
 			if err != nil {
-				log.Debugf("[Warning] failed to convert port value:%q to int with error: %+v", port, err)
+				return err
```
loicalbertin: This error should be Wrapped

PR #20, opened 2018-04-11
Remove not necessary job_name parameter and potentially not correctly…

stebenoist:
## Description of the change

- Remove job_name parameter in slurm squeue command because job_id is sufficient for slurm to identify an unique job and job_name can contains unescaped characters for slurm CLI
- Refactor some code
- Add Tests

### How to verify it

Deploy a Compute node on Slurm location

### DIFF
--- a/prov/slurm/helper.go
+++ b/prov/slurm/helper.go
@@ -130,8 +130,8 @@ func getAttributes(client sshutil.Client, key string, params ...string) ([]strin
 			if len(split) != 2 {
 				return nil, errors.Errorf("Slurm returned an unexpected stdout: %q with command:%q", out, cmd)
 			}
-			node := strings.Trim(split[0], "\" \t\n")
-			part := strings.Trim(split[1], "\" \t\n")
+			node := strings.Trim(split[0], "\" \t\n\x00")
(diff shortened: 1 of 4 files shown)


prov/slurm/helper.go
```diff
@@ -102,26 +102,42 @@ func checkInfraConfig(cfg config.Configuration) error {
...
+			out, err := client.RunCommand(cmd)
+			if err != nil {
+				return nil, errors.Wrap(err, "Failed to retrieve Slurm node name/partition")
+			}
+			split := strings.Split(out, ",")
+			if len(split) != 2 {
+				return nil, errors.Errorf("Slurm returned an unexpected stdout: %q with command:%q", out, cmd)
+			}
+			node := strings.Trim(split[0], "\" \t\n")
+			part := strings.Trim(split[1], "\" \t\n")
```
laurentganne: Add null character trimming too in both lines above ?
node := strings.Trim(split[0], "\" \t\n\x00")
part := strings.Trim(split[1], "\" \t\n\x00")

laurentganne: Ah no sorry, done in runCommand

stebenoist: Yes, it can be useful too

PR #21, opened 2018-04-19
Bugfix/brbdcf 1227 yorc too many open files

stebenoist:
## Description of the change
1. Add a ConsulHTTPClient as a singleton
2. Improve performance with ConsulHTTPClient by increasing MaxIdleConnsPerHost and MaxIdleConns in order to reuse existing connections instead of recreate new ones.
3. Add a SSH sessions pool to ensure only one SSH connection is used for any SSH client. This should fix the "too many open files" issue.

### How to verify it
1. Check the updated "no-op" deployment bench.
2. Deploy/Undeploy a single compute many times on Slurm location and check the SSH connection is unique.


stebenoist:
Yes, we need updating yorc vendor on plugins.

PR #23, opened 2018-05-02
Feature/brbdcf 610 monitor compute liveness

stebenoist:
## Description of the change
- Add pre and post activity hook to handle compute monitoring for delegate/setState operations.
- Create/Remove checks to make TCP SSH connection on computes at defined time interval stored in Consul KV Store
- Create a monitoring manager responsible for polling and starting checks when yorc node is leader on the monitoring service (https://www.consul.io/docs/guides/leader-election.html)
- Create events and logs on compute state when connection is on failure and when it's back to normal.
### How to verify it
- Install a Yorc cluster with at least 2 Yorc nodes to start/stop leader instance and validate failover.
Follow the doc (http://yorc.readthedocs.io/en/latest/ha.html#alien4cloud to configure Alien with HA
- Deploy one or many computes on any infrastructure like AWS or OpenStack with the orchestrator property **monitoring_time_interval** set to 5 seconds (by example) in Alien4Cloud
- Stop a compute and check its state passing to error (red in Alien4Cloud) with corresponding logs.
- Restart it and check its state is back to normal (greenin Alien4Cloud) with corresponding logs.
You can check Consul UI too to see health checks.
- Stop and restart Yorc server to check monitoring is still managed


stebenoist:
Yes, I've just fix it

PR #42, opened 2018-07-03
Feature/brbdcf 1356 deployments latency

laurentganne:
# Pull Request description

## Description of the change

Added Yorc server configuration parameters to improve the performance of bash/python scripts implementation artifacts when needed.

### What I did

Added 2 new configuration parameters in the Ansible section of Yorc server configuration :

- **cache_facts**: If set to true, caches Ansible facts (values fetched on remote hosts about network/hardware/OS/virtualization configuration) so that these facts are not recomputed each time a new operation is a run for a given deployment (false by default: no caching).
- **archive_artifacts**: If set to true, archives operation bash/python scripts locally, copies this archive and unarchives it on remote hosts (requires tar to be installed on remote hosts), to avoid multiple time consuming remote copy operations of individual scripts (false by default: no archive).

Updated the Yorc Server configuration documentation to describe these new parameters.
In the Ansible Configuration section, added a paragraph on performance considerations, to describe ansible playbooks performs better than scripts artifacts, and to describe parameters that can be used to improve performances when scripts are used.

### How to verify it

First, keeping default values in Yorc Server configuration, upload the Consul Component in Ystia Forge at https://github.com/ystia/forge/tree/develop/org/ystia/consul
Create a topoloy with one such Consul component on a compute node, and deploy it on an OpenStack location with a CentOS 7 on-demand compute node.
Check in yorc deployment logs, that the consul deployment (once the Compute node has been created) takes more than 5 minutes.

Undeploy.

Then change the Yorc server configuration to have these Ansible settings :

>   "ansible": {
>     "use_openssh": true,
>     "debug": true,
>     "keep_operation_remote_path": true,
>     "archive_artifacts": true,
>     "cache_facts": true
>   }

These 3 parameters are the parameters allowing to improve performances: use_openssh, archive_artifacts, cache_facts.

Restart the Yorc Server with this new config, and deploy again the Consul topology.
Check the consul deployment (once the Compute node has been created) takes less than 1 minute now.

### DIFF
--- a/commands/server.go
+++ b/commands/server.go
@@ -223,7 +223,7 @@ func setConfig() {
 	serverCmd.PersistentFlags().Int("ansible_connection_retries", 5, "Number of retries in case of Ansible SSH connection failure")
 	serverCmd.PersistentFlags().String("operation_remote_base_dir", ".yorc", "Name of the temporary directory used by Ansible on the nodes")
 	serverCmd.PersistentFlags().Bool("keep_operation_remote_path", config.DefaultKeepOperationRemotePath, "Define wether the path created to store artifacts on the nodes will be removed at the end of workflow executions.")
-	serverCmd.PersistentFlags().Bool("ansible_archive_artifacts", config.DefaultArchiveArtifacts, "Define wether artifacts should be archived before being copied on remote nodes.")
+	serverCmd.PersistentFlags().Bool("ansible_archive_artifacts", config.DefaultArchiveArtifacts, "Define wether artifacts should be archived before being copied on remote nodes (requires tar to be installed on remote nodes).")
 	serverCmd.PersistentFlags().Bool("ansible_cache_facts", config.DefaultCacheFacts, "Define wether Ansible facts (useful variables about remote hosts) should be cached.")
 
 	//Bind Consul persistent flags
--- a/commands/server.go
+++ b/commands/server.go
@@ -47,6 +47,8 @@ var ansibleConfiguration = map[string]interface{}{
 	"ansible.connection_retries":         5,
 	"ansible.operation_remote_base_dir":  ".yorc",
 	"ansible.keep_operation_remote_path": config.DefaultKeepOperationRemotePath,
+	"ansible.archive_artifacts":          config.DefaultArchiveArtifacts,
+	"ansible.fact_caching":               config.DefaultFactCaching,
 }
 
 var consulConfiguration = map[string]interface{}{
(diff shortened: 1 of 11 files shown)


laurentganne:
Note: build failures are not related to this pull request, they are du to a change in external go stringer behavior regarding dependencies. This change will be addressed in a separate pull request. In the meantime, the following command can be run before building Yorc : go install -i ./events

commands/server.go
```diff
@@ -221,6 +223,8 @@ func setConfig() {
 	serverCmd.PersistentFlags().Int("ansible_connection_retries", 5, "Number of retries in case of Ansible SSH connection failure")
 	serverCmd.PersistentFlags().String("operation_remote_base_dir", ".yorc", "Name of the temporary directory used by Ansible on the nodes")
 	serverCmd.PersistentFlags().Bool("keep_operation_remote_path", config.DefaultKeepOperationRemotePath, "Define wether the path created to store artifacts on the nodes will be removed at the end of workflow executions.")
+	serverCmd.PersistentFlags().Bool("ansible_archive_artifacts", config.DefaultArchiveArtifacts, "Define wether artifacts should be archived before being copied on remote nodes.")
```
loicalbertin: Can we explicitly say that tar is required on remote hosts here?

doc/configuration.rst
```diff
@@ -316,6 +332,25 @@ All available configuration options for Ansible are:
 
       * ``env``: An optional list environment variables to set when creating the container. The format of each variable is ``var_name=value``.
 
+
+Ansible performance considerations
```
loicalbertin: @laurentganne @adidanes @stebenoist What do you think about having a dedicated page about performance considerations?
Could be later enriched with others considerations.

commands/server.go
laurentganne: yes better to have it here as well, not just in the documentation

doc/configuration.rst
laurentganne: Maybe there aren't enough materials for now, would a page just containing the same details as above be ok ?  It would give:

# Performance

## TOSCA Operations

As described in TOSCA :ref:`tosca_operations_implementations_section`, Yorc supports these builtin implementations for operations to execute on remote hosts :

  * Bash scripts
  * Python scripts
  * Ansible Playbooks

It is recommended to implement operations as Ansible Playbooks to get the best execution performance.

When operations are not implemented using Ansible playbooks, the following Ansible configuration settings allow to improve the performance of scripts execution on remote hosts :

  * `use_openssh`: Prefer OpenSSH over Paramiko, a Python implementation of SSH (used by default) to provision remote hosts. OpenSSH have several optimization like reusing connections that should improve preformance but may lead to issues on older systems See Ansible documentation on Remote connection information.
  * `cache_facts`: Caches Ansible facts (values fetched on remote hosts about network/hardware/OS/virtualization configuration) so that these facts are not recomputed each time a new operation is a run for a given deployment.
  * `archive_artifacts`: Archives operation bash/python scripts locally, copies this archive and unarchives it on remote hosts (requires tar to be installed on remote hosts), to avoid multiple time consuming remote copy operations of individual scripts.

loicalbertin: @adidanes @stebenoist WDYT?

stebenoist: +1 for this specific perf paragraph

commands/server.go
stebenoist: We could install tar on remote hosts if not present and if the archive_artifacts option has been selected ?

stebenoist: That's ok for me.

PR #43, opened 2018-07-04
Bugfix/replace stringer

stebenoist:
# Pull Request description

Replace last uses of stringer generation code by go-enum


deployments/structs.go
```diff
@@ -14,29 +14,16 @@
...
+// DeploymentStatus x ENUM(
+// INITIAL,
+// DEPLOYMENT_IN_PROGRESS,
+// DEPLOYED,
+// UNDEPLOYMENT_IN_PROGRESS,
+// UNDEPLOYED,
+// DEPLOYMENT_FAILED,
+// UNDEPLOYMENT_FAILED,
+// SCALING_IN_PROGRESS
+// )
```
laurentganne: Minor issue but go convention is to use MixedCaps instead of underscores, and code climate is raising issues on this.
As these enums will be prefixed by DeploymentStatus, could we have somzthing like :

Initial,
InProgress,
Deployed,
Undeploying,
Undeployed,
Failed,
UndeploymentFailed,
ScalingInProgress

stebenoist: Yes I agree

stebenoist: But... it impacts the generated string method. I see what is possible.

stebenoist: I just can remove the DeploymentStatus prefix as before. WYT ?

laurentganne: ok

PR #51, opened 2018-07-11
Feature/add doc around sshpass

stebenoist:
- Add setup note about the need to install sshpass if hosts pool are used with password authentication
- Fix minor issue about "http: multiple response.WriteHeader calls"

### DIFF
--- a/doc/install.rst
+++ b/doc/install.rst
@@ -63,8 +63,7 @@ softwares should also be installed.
   
   sudo pip install docker-py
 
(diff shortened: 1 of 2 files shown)


doc/install.rst
```diff
@@ -63,6 +63,16 @@ softwares should also be installed.
   
   sudo pip install docker-py
 
+To support :ref:`Hosts Pool <yorc_infras_hostspool_section>` with SSH password authentication instead of common ssh keys, you need
```
laurentganne: Isn't it applicable to all infrastructures using ansible ?

stebenoist: yes

Issue #72, opened 2018-07-11
Investigate why we can't build on go 1.11beta1


loicalbertin:
# Issue Type

Put an X between the brackets of the corresponding issue type

* [X] Bug report
* [ ] Documentation issue report
* [ ] Improvement request

## Description

See build https://travis-ci.org/ystia/yorc/builds/397266821?utm_source=github_status&utm_medium=notification

 
```
--> Running go test
# github.com/ystia/yorc/vendor/github.com/armon/go-metrics/prometheus
vendor/github.com/armon/go-metrics/prometheus/prometheus.go:1: +build comment must appear before package clause and be followed by a blank line%!(EXTRA []interface {}=[])
?   	github.com/ystia/yorc	0.058s	coverage: 0.0% of statements [no test files]
ok  	github.com/ystia/yorc/bench	0.057s	coverage: 0.0% of statements [no tests to run]
ok  	github.com/ystia/yorc/commands	0.087s	coverage: 67.5% of statements
?   	github.com/ystia/yorc/commands/deployments	0.061s	coverage: 9.7% of statements [no test files]
?   	github.com/ystia/yorc/commands/deployments/tasks	0.063s	coverage: 4.6% of statements [no test files]
?   	github.com/ystia/yorc/commands/deployments/workflows	0.060s	coverage: 13.9% of statements [no test files]
?   	github.com/ystia/yorc/commands/hostspool	0.056s	coverage: 9.8% of statements [no test files]
?   	github.com/ystia/yorc/commands/httputil	0.008s	coverage: 0.0% of statements [no test files]
ok  	github.com/ystia/yorc/config	0.006s	coverage: 78.4% of statements
# github.com/ystia/yorc/deployments
deployments/instances.go:141: Wrapf call needs 2 args but has 3 args
FAIL	github.com/ystia/yorc/deployments [build failed]
ok  	github.com/ystia/yorc/events	2.209s	coverage: 69.6% of statements
ok  	github.com/ystia/yorc/helper/collections	0.003s	coverage: 100.0% of statements
?   	github.com/ystia/yorc/helper/consulutil	0.005s	coverage: 0.0% of statements [no test files]
?   	github.com/ystia/yorc/helper/executil	0.003s	coverage: 0.0% of statements [no test files]
?   	github.com/ystia/yorc/helper/labelsutil	0.003s	coverage: 0.0% of statements [no test files]
[...]

PR #109, opened 2018-07-12
Bugfix/fix go1.11beta1 build issues

stebenoist:
- Fix new build issues for go1.11beta1 around string formatting
- Ignore vendor folder during go test cmd

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: go
 go:
-  - 1.x
+  - stable
 
 sudo: required
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: go
 go:
-  - 1.10.x
+  - 1.x
 
 sudo: required
 
--- a/Makefile
+++ b/Makefile
@@ -47,11 +47,11 @@ dist: buildnformat
 test: generate
 ifndef SKIP_TESTS
 	@echo "--> Running go test"
-	@export PATH=$$PWD/build:$$PATH; go test $(TESTARGS) -p 1 ./... | grep -v vendor
+	@export PATH=$$PWD/build:$$PATH; go test $(TESTARGS) -p 1 `go list ./...`
 endif
 
 
-cover: 
+cover:
 	@go test -p 1 -cover $(COVERARGS) ./...  
 
 format:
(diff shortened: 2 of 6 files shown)


prov/validation/compute_capability_endpoint.go
```diff
@@ -35,7 +35,7 @@ func postComputeCreationHook(ctx context.Context, cfg config.Configuration, task
 	cc, err := cfg.GetConsulClient()
 	if err != nil {
 		events.WithContextOptionalFields(ctx).NewLogEntry(events.LogLevelWARN, deploymentID).
-			Registerf("Failed to retrieve consul client when ensuring that a compute will have it's endpoint ip set. Next operations will likely failed: %v", target, err)
+			Registerf("Failed to retrieve consul client when ensuring that a compute will have it's endpoint ip set. Next operations will likely failed: %v", err)
```
laurentganne: typo in already existing code: "will likely failed" instead of "willl likely fail"

Makefile
```diff
@@ -47,7 +47,7 @@ dist: buildnformat
 test: generate
 ifndef SKIP_TESTS
 	@echo "--> Running go test"
-	@export PATH=$$PWD/build:$$PATH; go test $(TESTARGS) -p 1 ./...
+	@export PATH=$$PWD/build:$$PATH; go test $(TESTARGS) -p 1 ./... | grep -v vendor
```
loicalbertin: From my understanding it only exclude vendor from test output, not from actually testing the vendor directory. 
The way your are doing it will result on grep command returning always a 0 return code even if our tests are failing.

I think that the result of `go list ./...` would do the job.

Did I miss something

.travis.yml
```diff
@@ -1,6 +1,6 @@
 language: go
 go:
-  - 1.10.x
+  - 1.x
```
loicalbertin: Is there a way to exclude pre-releases?

Makefile
stebenoist: Hum yes I missed it

.travis.yml
stebenoist: "stable" should do the job as travis use gimme
https://github.com/travis-ci/gimme
I'll try it...

loicalbertin:
Closes #72

Issue #76, opened 2018-07-11
Deploy Kubernetes Application in a specific namespace instead of creating a new one


loicalbertin:
As: An Kubernetes operator

I want to: Impose that Kubernetes apps deployed by Yorc use a specific namespace

So that: I have my own reasons! I'm an OPS! (Security, auth, limits, monitoring,...)

 

AC1: Should be configurable 

AC2: Idealy configurable in A4C to allow mapping.

AC3: If no namespace specified, then create a new one as for now

PR #133, opened 2018-07-20
Feature/deploy k8s namespace

adidanes:
# Pull Request description
If a namespace is provided in the k8s resources specification (in the associated metadata), Yorc have to create the resources in that namespace. If not provided, Yorc creates a namespace at deployment for all the created resources, and deletes the namespace after the resources are deleted at undeployment. 

Resolves issue : https://github.com/ystia/yorc/issues/76

### What I did
Two functions in k8_utils.go manage k8s namespaces.
- defaultNamespace() creates a namespace for a deployment as before
- getNamespace() returns the namespace provided in the resources specification, and if not provided, returns the result of defaultNamespace()

### How I did it
I suppose that if the namespace is provided in the resources metadata, the following is true:
- the provided namespace pre-exists

### How to verify it
Make 2 deployments:

1.  As before (no namespace defined in the resources metadata) ; check that:
- [ ]  a k8s namespace is created (the deploymentId is used as before)
- [ ] the created k8s objects belong to this namespace
- [ ] after undeployment, the namespace and the objects are deleted

2. create a zip by modifying the previous one and adding "namespace" metadata to the resources specifications ; the value of the "namespace" metadata must correspond to a pre-existent namespace on the k8s cluster used to deploy ; check that:

- [ ] created k8s objects are created in the specified namespace
- [ ] after the undeploy, objects are deleted but namespace still exists

Fragment of the modified topology:
resource_spec: |
            {
              "metadata" : {
                "name" : "containerdeployment--1764703054",
                **"namespace": "myapache"**
              },
              "apiVersion" : "apps/v1beta1",
              "kind" : "Deployment",
              ...

## Verify the complete chain (Alien -> Yorc -> K8S cluster)

This corresponds to HowTo Demo.

### Setup environment

* tested with a SNAPSHOT of alien4cloud 2.1.0 : http://10.197.135.244/starlings/alien4cloud/alien4cloud-premium-dist-2.1.0-20180710.004211-8-dist.tar.gz
* alien4cloud-kubernetes-plugin must be generated for this branch : https://github.com/alien4cloud/alien4cloud-kubernetes-plugin/tree/feature/support_k8s_namespaces
* alien4cloud-yorc-plugin : https://github.com/ystia/yorc-a4c-plugin/tree/feature/k8s_namespaces
* add to alien Catalog built CSAR for Apache ; can also add sample topology simple-apache (or add Git Location with Repository URL https://github.com/ystia/yorc-a4c-plugin.git, branch feature/k8s_namespaces and archives to import tosca-samples/org/ystia/yorc/samples/kube)
* create Yorc orchestrator and Kubernetes location
* Yorc must be configured with infrastructure definition for an existing, pre-deployed k8s cluster.
* create a meta-property for the k8s namespace

## Create test application

* Use the Apache component and simple-apache template
* Set created meta-propery's value to an existing namespace Name from your k8s cluster

## Verify
- [ ] Deploy application and check that k8s objects (deployment, replica set, service) were created in the specified namespace
- [ ] Undeploy application and check that previously created k8s objects (deployment, replica set, service) no longer exist in the specified namespace


prov/kubernetes/k8s_utils.go
```diff
@@ -170,3 +170,29 @@ func referenceFromObjectReference(ref corev1.ObjectReference) reference {
...
+// Return it if provided ; return one generated with the default policy, if not provided
+func getNamespace(deploymentID string, objectMeta metav1.ObjectMeta) (string, bool) {
+	var isProvided bool
+	var namespace string
+	var providedNamespace string
+	if &objectMeta != nil {
+		providedNamespace = objectMeta.Namespace
+	}
+
+	if &providedNamespace != nil && providedNamespace != "" {
```
loicalbertin: `&providedNamespace != nil` seems unnecessary & over-complicated `providedNamespace != ""` is enough IMHO

prov/kubernetes/k8s_utils.go
```diff
@@ -170,3 +170,29 @@ func referenceFromObjectReference(ref corev1.ObjectReference) reference {
...
+		namespace = providedNamespace
+		isProvided = true
+	} else {
+		namespace, _ = defaultNamespace(deploymentID)
+		isProvided = false
```
loicalbertin: Unnecessary: the zero value for a `bool` is already `false`

prov/kubernetes/execution.go
```diff
@@ -275,14 +277,15 @@ func (e *executionCommon) manageDeploymentResource(ctx context.Context, clientse
 
 		events.WithContextOptionalFields(ctx).NewLogEntry(events.LogLevelINFO, e.deploymentID).Registerf("k8s Deployment %s deleted", deploymentName)
 
-		// Delete namespace
-		err = generator.deleteNamespace(namespace, clientset)
-		if err != nil {
-			events.WithContextOptionalFields(ctx).NewLogEntry(events.LogLevelINFO, e.deploymentID).Registerf("Cannot delete %s k8s Namespace", namespace)
-			return err
+		// Delete namespace if not provided
```
loicalbertin: Maybe not related to this US but what if we have several deployments in the same app?
This is generally the case. Like in the monitoring sample. We will wait for the deletion of graphite then try to delete the namespace while grafana's deployment is still there.

adidanes: Maybe the same namespace should be used for all the related deployments (graphite and grafana in the monitoring sample).

adidanes: I mean this problem does not arrive if we use the current new feature that allows to specify the namespace for all the k8s objects of the same app

loicalbertin:
README update is missing

loicalbertin:
LGTM

Issue #93, opened 2018-07-11
Split workflow execution unit to step


loicalbertin:
As: Yorc engineering team
I want to: Make the step the task unit of a workflow execution
So that: it will be the first step to distributed deployments

AC1: Different steps of a same workflow could be executed on different Yorc instances
AC2: Yorc instances synchronize themselves using consul to know which one will handle a step
AC3: Reuse as much as possible the distributed tasks framework already implemented in Yorc

 

Use case:

Given 2 instances of Yorc with 1 worker on each.

And a application composed of just 2 computes.

I want to see deployment of 2 computes happening in parallel

PR #135, opened 2018-07-27
Feature/split workflow exec into step

stebenoist:
# Pull Request description
## Description of the change
This PR implements #93 

It allows to define a workflow step as unit execution task in order to split the workflow execution into different workers of one or many Yorc instances

### What I did
Define a task as an execution for non workflow task
Define a task as n executions for workflow task with n equal to the steps number
#### 1. Refactor the tasks Collector as to define Collector concerns:
- register in Consul the task (/tasks/)
- register in Consul the workflow if task is workflow (/workflows/)
- register in Consul the executions for initial steps if task is workflow (/executions/)
- register in Consul the execution for task if non workflow task
#### 2. Just change the dispatcher to handle task executions instead of tasks
Dispatcher concern is polling executions task and dispatch them across available workers
- It has to acquire a lock on the execution task as other distributed dispatchers can try to do the same
- If it gets the lock, it instantiates an execution task and push it to workers pool
- If it receives a message from shutdown channel, it has to spread the shutdown to workers
#### 3. Changes the worker concern to execute workflow step as execution instead of whole workflow
- worker executes a task execution sent from the dispatcher
- It has to poll cancellation and failure flags to stop execution
-  In case of workflow task, it runs a workflow step as a task execution and needs to update related task status, and register next step tasks execution if its execution is successful. In case of workflow join, it needs to check all previous steps are done.
- It has to release related execution Consul lock and mark the execution to delete
- If workflow is done, the task and deployment statuses need to be updated
- If an error occurred during the step execution, task status need to be updated
- Each done step will register the next ones. In case of join, the last done of previous steps will register the next step
- For a final step, it checks the status of all steps and if all are done, it sets the task status to done

### How to verify it
As this impacts many things in orchestrator, non-regression tests need to be run with all task types as custom command, custom workflow, query, deploy, undeploy, purge, cancel, resume in distributed and non-distributed modes.
To ensure, the goal is reached, a topology with 2 computes has to be deployed by a Yorc cluster with 2 nodes, each node with only one worker so that each node runs the install of one compute as these steps are parallel.

## Applicable Issues
#93


adidanes:
Why after code refactoring, the _Dispatcher_ and the _worker_ interfaces are kept in the _workflow_ package ?
The Dispatcher and the Workers now are also concerned by tasks that are not related to workflow step execution.

loicalbertin:
My main concern about this PR is that it totally breaks [the compatibility promise](https://golang.org/doc/go1compat). This should be avoided outside a major release.

commands/deployments/tasks/dep_fix_task_step.go
```diff
@@ -49,7 +49,7 @@ var updateTaskStepCmd = &cobra.Command{
 		}
 
 		// The task step status is set to "done"
-		step := &tasks.TaskStep{Name: args[2], Status: strings.ToLower(tasks.TaskStepStatusDONE.String())}
+		step := &tasks.TaskStep{Name: args[2], Status: strings.ToLower(tasks.StepStatusDONE.String())}
```
loicalbertin: Compat is broken

tasks/collector.go
```diff
@@ -1,133 +0,0 @@
...
-
-// NewCollector creates a Collector
-func NewCollector(consulClient *api.Client) *Collector {
-	return &Collector{consulClient: consulClient}
-}
-
-// RegisterTaskWithData register a new Task of a given type with some data
-//
-// The task id is returned.
-func (c *Collector) RegisterTaskWithData(targetID string, taskType TaskType, data map[string]string) (string, error) {
```
loicalbertin: Compat broken

tasks/collector.go
```diff
@@ -1,133 +0,0 @@
...
-		return "", err
-	}
-	return taskID, nil
-}
-
-// RegisterTask register a new Task of a given type.
-//
-// The task id is returned.
-// Basically this is a shorthand for RegisterTaskWithData(targetID, taskType, nil)
-func (c *Collector) RegisterTask(targetID string, taskType TaskType) (string, error) {
```
loicalbertin: Compat broken

tasks/structs_step.go
```diff
@@ -16,16 +16,16 @@ package tasks
...
-// TaskStepStatus x ENUM(
+// StepStatus x ENUM(
 // INITIAL,
 // RUNNING,
 // DONE,
 // ERROR,
 // CANCELED
 // )
-type TaskStepStatus int
+type StepStatus int
```
loicalbertin: Compat broken

tasks/structs_step_enum.go
```diff
@@ -23,52 +23,52 @@ import (
...
+	_StepStatusName[23:31]:                  4,
+	strings.ToLower(_StepStatusName[23:31]): 4,
 }
 
-// ParseTaskStepStatus attempts to convert a string to a TaskStepStatus
-func ParseTaskStepStatus(name string) (TaskStepStatus, error) {
-	if x, ok := _TaskStepStatusValue[name]; ok {
-		return TaskStepStatus(x), nil
+// ParseStepStatus attempts to convert a string to a StepStatus
+func ParseStepStatus(name string) (StepStatus, error) {
```
loicalbertin: Compat broken

tasks/tasks.go
```diff
@@ -32,6 +32,36 @@ import (
...
+	targetID string
+	status   string
+}
+
+func (e anotherLivingTaskAlreadyExistsError) Error() string {
+	return fmt.Sprintf("Task with id %q and status %q already exists for target %q", e.taskID, e.status, e.targetID)
+}
+
+// NewAnotherLivingTaskAlreadyExistsError allows to create a new anotherLivingTaskAlreadyExistsError error
+func NewAnotherLivingTaskAlreadyExistsError(taskID, targetID, status string) error {
```
loicalbertin: Sounds weird to do this. This means that `anotherLivingTaskAlreadyExistsError` should live in the collector package. But this will break the compatibility promise.

tasks/workflow/step.go
```diff
@@ -0,0 +1,342 @@
...
+	"github.com/ystia/yorc/registry"
+	"github.com/ystia/yorc/tasks"
+	"github.com/ystia/yorc/tosca"
+	"path"
+	"strings"
+	"time"
+)
+
+// Step represents the workflow step
+type Step struct {
```
loicalbertin: Is there a reason to make `Step` public? I didn't see any for now.

tasks/workflow/task_execution.go
```diff
@@ -0,0 +1,122 @@
...
+	"fmt"
+	"github.com/hashicorp/consul/api"
+	"github.com/pkg/errors"
+	"github.com/ystia/yorc/helper/consulutil"
+	"github.com/ystia/yorc/log"
+	"github.com/ystia/yorc/tasks"
+)
+
+// A TaskExecution is the unit task of execution. If task is a workflow, it contains as many TaskExecutions as workflow steps
+type TaskExecution struct {
```
loicalbertin: Is there any reason to make `TaskExecution` public? I can't see any for now.

loicalbertin:
README update is missing

PR #165, opened 2018-09-18
Feature/upgrade tf

stebenoist:
# Pull Request description
Upgrade Terraform version to latest stable 0.11.8

## Description of the change

- Upgrade Terraform
- Add --terraform_plugins_dir server option to define the directory where to find Terraform pre-installed providers plugins. If no specified, required plugins will be downloaded during deployment
- Upgrade documentation
- Upgrade Dockerfile to ship pre-installed Terraform plugins in /var/terraform/plugins
- Update google Terraform executor to match with new Terraform Google plugin new version


doc/configuration.rst
```diff
@@ -74,6 +74,10 @@ Globals Command-line options
 
   * ``--consul_ssl_verify``: If set to false, disable Consul certificate checking (true by default is ssl enabled).
 
+.. _option_terraform_plugins_dir_cmd:
+
+  * ``--terraform_plugins_dir``: Specify the directory where to find Terraform pre-installed providers plugins. If no specified, required plugins will be downloaded during deployment. See https://www.terraform.io/guides/running-terraform-in-automation.html#pre-installed-plugins for more information.
```
laurentganne: typo: If no specified -> If not specified

laurentganne: In addition to a new section in CHANGELOG.md to describe the Yorc dependency on terraform version and constraints on plugins versions (see similar dependency section in changelog in 3.1.0-M2 for ansible),
Maybe we could add here as well the plugin versions constraints Yorc has on plugins, else the user would discover only later at usage he has downloaded a plugin version incompatible with Yorc hardcoded expectations

config/config.go
```diff
@@ -61,6 +61,18 @@ const DefaultCacheFacts = false
 // DefaultWfStepGracefulTerminationTimeout is the default timeout for a graceful termination of a workflow step during concurrent workflow step failure
 const DefaultWfStepGracefulTerminationTimeout = 2 * time.Minute
 
+// DefaultTFConsulPluginVersionConstraint is the default Terraform Consul plugin version
+const DefaultTFConsulPluginVersionConstraint = "~> 2.1"
```
loicalbertin: Why not setting this from versions.yaml?
Sounds like a duplicate

doc/configuration.rst
```diff
@@ -74,6 +74,26 @@ Globals Command-line options
 
   * ``--consul_ssl_verify``: If set to false, disable Consul certificate checking (true by default is ssl enabled).
 
+.. _option_terraform_plugins_dir_cmd:
+
+  * ``--terraform_plugins_dir``: Specify the directory where to find Terraform pre-installed providers plugins. If not specified, required plugins will be downloaded during deployment. See https://www.terraform.io/guides/running-terraform-in-automation.html#pre-installed-plugins for more information.
+
+.. _option_terraform_aws_plugin_version_constraint_cmd:
+
+  * ``--terraform_aws_plugin_version_constraint``: Specify the Terraform AWS plugin version constraint. Default one compatible with our source code is ``"~> 1.36"``. If you choose another, it's at your own risk. See https://www.terraform.io/docs/configuration/providers.html#provider-versions for more information.
```
loicalbertin: Same as previous comment we may be able to automate generation of those versions by looking at the versions.yaml file. 
versions.yaml should be the only source of truth.

prov/terraform/commons/resources.go
```diff
@@ -25,6 +25,8 @@ const (
 	// DefaultSSHPrivateKeyFilePath is the default SSH private Key file path
 	// used to connect to provisioned resources
 	DefaultSSHPrivateKeyFilePath = "~/.ssh/yorc.pem"
+	// NullPluginVersionConstraint is the Terraform null plugin version constraint
+	NullPluginVersionConstraint = "~> 1.0"
```
loicalbertin: Why this is defined here while others plugin are in config?

versions.yaml
```diff
@@ -1,4 +1,8 @@
 yorc_version: 3.1.0-SNAPSHOT
 consul_version: 1.0.6
-terraform_version: 0.9.11
+terraform_version: 0.11.8
 ansible_version: 2.6.3
+tf_consul_plugin_version: 2.1.0
+tf_aws_consul_plugin_version: 1.36.0
+tf_openstack_plugin_version: 1.9.0
+tf_google_plugin_version: 1.18.0
```
loicalbertin: New line missing

prov/terraform/commons/resources.go
stebenoist: I guess this plugin is not supposed to change often.

loicalbertin: ok

PR #169, opened 2018-09-26
WIP Feature/gh 130 yorc upgrade

### LINKED ISSUE #130: Upgrade From a Yorc version to another
As: An op
I want to: upgrade my yorc system and keep deployed envs
So that: I can easily benefit from Yorc improvements & bugfixes

AC1: Trace schema
AC2: update topology + consul schema

loicalbertin:
# Pull Request description

## Description of the change

Setup a mechanism that trace Consul database schema version and provide a way to migrate from one version to another.

### What I did

* Trace schema version in Consul.
* At startup Yorc take a lock and check its supported version against the current database version
  * If its version is upper than the current version this instance proceed with the schema upgrade (nothing done currently). And finally it updates the current schema version in Consul
  * If its version is lower than the current version this instance stops with a corresponding error message

Allow to upgrade from version 3.0.0 to a newer version without loosing existing data ([GH-130](https://github.com/ystia/yorc/issues/130))

## Applicable Issues

Fixes #130

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,6 +21,7 @@ work
 **/debug
 build/consul
 /doc/_build/*
+/doc/.doctrees
 /dist
 /plugin/testdata/plugin-sample/plugin-sample
 build/coverprofiles/*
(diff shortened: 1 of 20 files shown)


adidanes:
Concerning the **upgradeFromVersion** function, ligne https://github.com/ystia/yorc/blob/feature/GH-130-yorc_upgrade/server/consul.go#L131,
- should change _to old_ with _too old_
- maybe we could specify the current DB schema version, and which Yorc version the user could install (starting with that version the used DB schema is equal with, or greater then, the one registered in Consul)

loicalbertin:
> Concerning the **upgradeFromVersion** function, ligne https://github.com/ystia/yorc/blob/feature/GH-130-yorc_upgrade/server/consul.go#L131,
> 
> * should change _to old_ with _too old_

Fixed thx.

> * maybe we could specify the current DB schema version, and which Yorc version the user could install (starting with that version the used DB schema is equal with, or greater then, the one registered in Consul)

We can't know in advance the version of Yorc that will ship a given version of the schema. Those versions are not correlated.
But It's a good idea to print the schema version using the version command.

PR #173, opened 2018-10-04
Feature/gcloud static ip

### LINKED ISSUE #82: Support GCE Public IPs
As: an app manager
I want to: Access GCE computes from a public IP
So that: expose services

AC1: Could provision a new Public IP and associate it to a compute
AC2: Could reuse an existing Public IP and associate it to a compute

stebenoist:
# Pull Request description
Allow to deploy a compute on GCE with an external static IP.
AC1: Could provision a new Public IP and associate it to a compute
AC2: Could reuse an existing Public IP and associate it to a compute

### What I did
- Remove **yorc.nodes.google.Compute** _addresses_ property
- Tosca modelisation for Google Address node type **yorc.nodes.google.Address** wich owns the **yorc.capabilities.Assignable** capability
- Tosca Modelisation for operation artifact type **yorc.artifacts.google.Deployment.Address** which allows to run "**create**" and "**delete**" operations as to assign/unassign an external Address to a compute.
- Register TerraformGoogle Operation executor to execute operations with **yorc.artifacts.google.Deployment.Address** type

### How to verify it
**Here are many testcases**:
1. Create a topology with an google compute and the no_address property checked (means only ephemeral private address)
Install yorc on a GCP zone and deploy a Google Compute on the same zone with no_address property checked
Check deployment is OK.
Check compute attributes
Check with the gcloud CLI that the compute has a private IP address
For a private IP address, no EXTERNAL_IP address is present in the gcloud compute instances list

2. Create a topology with an abstract compute
Check deployment is OK.
Check compute attributes.
Check with the gcloud CLI that the compute has a ephemeral public IP address (it doesn't appear in the addresses list as it's not a static address)

3. Create a topology with an abstract compute and abstract network
Check topology modifications is OK at post location event
Check deployment is OK.
Check compute attributes
Check with the gcloud CLI that the compute has a static public IP address

4. Create a topology with an abstract compute and a concrete google public network
(command: gcloud compute addresses create my-static-ip)

Add the generated address ip in the related google public network node property
Execute the commands:
- gcloud compute instances list
- gcloud compute addresses list
For a static IP address, the compute must be present in the addresses list and its address ip in the computes list in the EXTERNAL_IP column

Idem with scalable compute -)

## Applicable Issues
fixes #82


data/tosca/yorc-google-types.yml
```diff
@@ -8,6 +8,10 @@ metadata:
 imports:
   - yorc: <yorc-types.yml>
 
+artifact_types:
+  yorc.artifacts.google.Deployment.Address:
```
loicalbertin: Just a question, I do not have a definitive position on  that but should we have a specific artifact type for each node type or a generic one like `yorc.artifacts.google.Deployment` used for all gcp resources ?

I think I prefer the second option as we will probably have some other nodes like this in future.
Do you see any drawbacks with the second option?
@laurentganne @adidanes WDYT?

deployments/definition_store.go
```diff
@@ -1129,11 +1140,10 @@ func createNodeInstances(consulStore consulutil.ConsulStore, kv *api.KV, numberI
...
 }
 
-/**
-This function check if a nodes need a floating IP, and return the name of Floating IP node.
-*/
-func checkFloattingIP(kv *api.KV, deploymentID, nodeName string) (bool, string, error) {
-	requirementsKey, err := GetRequirementsKeysByTypeForNode(kv, deploymentID, nodeName, "network")
+// This function check if a nodes need a specified requirement with a specified capacity type or derived
+// and return the name of the related node addressing the capacity
+func checkRequirements(kv *api.KV, deploymentID, nodeName, requirement, capabilityType string) (bool, string, error) {
```
loicalbertin: I think it is an interesting function that could be exposed as part of the `deployments` package api. We need to change it's name, move it to the requirements file and document it a bit more. WDYT?

prov/terraform/commons/resources.go
```diff
@@ -151,3 +151,37 @@ func GetConnInfoFromEndpointCredentials(kv *api.KV, deploymentID, nodeName strin
 	}
 	return user.RawString(), pkfp, nil
 }
+
+// HasAnyRequirement returns true and the the node name providing the capability
+// if node has a defined requirement type derived from the specified type or equal to the specified type
+func HasAnyRequirement(kv *api.KV, deploymentID, nodeName, derivedType, requirement string) (bool, string, error) {
```
loicalbertin: This function seems to have nothing specific to terraform. It should be part of the `deployments` package

prov/terraform/google/generator.go
```diff
@@ -117,34 +117,44 @@ func (g *googleGenerator) GenerateTerraformInfraForNode(ctx context.Context, cfg
...
-				// Do not generate something for this node instance (will be deleted if exists)
-				continue
+
+			err = g.generateComputeInstance(ctx, kv, cfg, deploymentID, nodeName, instanceName, instNb, &infrastructure, outputs)
+			if err != nil {
+				return false, nil, nil, err
 			}
-			err = g.generateComputeInstance(ctx, kv, cfg, deploymentID, nodeName, instanceName, index, &infrastructure, outputs)
+
+		case "yorc.nodes.google.PublicNetwork":
```
loicalbertin: Should we take care of this at all in yorc engine?
Those types should be defined only in the yorc-a4c-plugin and transformed into a `Address` by a modifier.

data/tosca/yorc-google-types.yml
stebenoist: Ok for the second option in my opinion

prov/terraform/google/generator.go
stebenoist: That's correct

prov/terraform/commons/resources.go
stebenoist: Ok

deployments/definition_store.go
stebenoist: Yes, this function seems to be very closed to the HasAnyRequirement in the prov terraform commons package you mentioned below

PR #176, opened 2018-10-12
Feature/google block storage

### LINKED ISSUE #81: Support GCE BlockStorages
As: an app manager
I want to: attach block storage on an GCE Compute
So that: I can persist informations on those persistent storages

AC1: Could provision a new volume and attach it to a compute
AC2: Could reuse an existing volume and attach it to a compute
AC3: Compute could have multiple volumes attached
AC4: All above should work with scalable computes

stebenoist:
# Pull Request description

**Support GCE Block Storages**

As: an app manager
I want to: attach block storage on an GCE Compute
So that: I can persist informations on those persistent storages

AC1: Could provision a new volume and attach it to a compute
AC2: Could reuse an existing volume and attach it to a compute
AC3: Compute could have multiple volumes attached
AC4: All above should work with scalable computes

### What I did
- Create yorc.nodes.google.PersistentDisk node type
- Create relationship **yorc.relationships.google.AttachesTo** with specific prop **mode** to allow
the possibility for a compute to attach a disk with **read only** mode.
- Create **scratch_disks** prop for google compute instance to attach local SSD disks.

### How I did it
- Use the existing code around topology modification for BlockStorage and compute **local_storage** 
requirement.

### How to verify it
- Simple topology with a scalable compute and one or many block storages.
You can use **yorc.relationships.google.AttachesTo** and set mode property to **READ_ONLY**
You can use generic **tosca.relationships.AttachTo**.
You can attach existing disk by setting **volume_id** property.
Check the deployment and the resources creation on gcloud console

## Applicable Issues
fixes #81

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## UNRELEASED
 
+### FEATURES
+
+* Support GCE Block storages. ([GH-82](https://github.com/ystia/yorc/issues/81))
+
 ## 3.1.0-M4 (October 08, 2018)
 
 ### DEPENDENCIES
--- a/data/tosca/yorc-google-types.yml
+++ b/data/tosca/yorc-google-types.yml
@@ -46,6 +46,12 @@ relationship_types:
         description: >
           The mode in which to attach this disk, either READ_WRITE or READ_ONLY. If not specified, the default is to attach the disk in READ_WRITE mode.
         required: false
+    attributes:
+      device:
+        type: string
+        description: >
+          The logical name of the device as exposed to the instance.
+          Note: A runtime property that gets set when the model gets instantiated by the orchestrator.
 
 node_types:
   yorc.nodes.google.Compute:
--- a/data/tosca/yorc-google-types.yml
+++ b/data/tosca/yorc-google-types.yml
@@ -25,6 +25,14 @@ data_types:
         required: false
         description: The RFC 4648 base64 encoded SHA-256 hash of the customer-supplied encryption key that protects this resource.
 
+  yorc.datatypes.google.ScratchDisk:
+    derived_from: tosca.datatypes.Root
+    properties:
+      interface:
+        type: string
+        required: false
+        description: The disk interface to use for attaching the scratch disks; either SCSI or NVME. Defaults to SCSI.
+
 node_types:
   yorc.nodes.google.Compute:
     derived_from: yorc.nodes.Compute
@@ -116,6 +124,15 @@ node_types:
           Comma-separated list of tags to apply to the instances for identifying
           the instances to which network firewall rules will apply.
         required: false
+      scratch_disks:
+        type: list
+        description: Additional scratch disks to attach to the instance. Maximum allowed is 8.
+        required: false
+        entry_schema:
+          type: yorc.datatypes.google.ScratchDisk
+          constraints:
+            - greater_or_equal: 0
+            - max_length: 8
     requirements:
       - assignment:
           capability: yorc.capabilities.Assignable
--- a/data/tosca/yorc-google-types.yml
+++ b/data/tosca/yorc-google-types.yml
@@ -225,32 +225,33 @@ node_types:
           Customer-supplied encryption keys do not protect access to metadata of the disk.
           If you do not provide an encryption key when creating the disk, then the disk will be encrypted using an automatically generated key and you do not need to provide a key to use the disk later.
         required: false
-      source_image:
+      image_id:
         type: string
         description: >
           The image from which to initialize this disk. This can be one of: the image's self_link, projects/{project}/global/images/{image}, projects/{project}/global/images/family/{family}, global/images/{image}, global/images/family/{family}, family/{family}, {project}/{family}, {project}/{image}, {family}, or {image}.
           If referred by family, the images names must include the family name. If they don't, use the google_compute_image data source. For instance, the image centos-6-v20180104 includes its family name centos-6. These images can be referred by family name here.
         required: false
-      source_image_encryption_key:
+      image_encryption_key:
         type: yorc.datatypes.google.EncryptionKey
         description: >
           The customer-supplied encryption key of the source image. Required if the source image is protected by a customer-supplied encryption key.
         required: false
-      source_snapshot:
+      snapshot_id:
         type: string
         description: >
           The source snapshot used to create this disk. You can provide this as a partial or full URL to the resource.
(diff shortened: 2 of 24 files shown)


laurentganne:
Could check a persistent disk was attached to the Compute instance,
however the Application template ConsulOnBS at https://github.com/ystia/forge/tree/develop/org/ystia/experimental/consul/topologies/ConsulOnBS can't be deployed on Google Cloud due to a difference of meaning for the device name on yorc.nodes.google.PersistentDisk, compared to what expects the Alien4Cloud types defined at https://github.com/alien4cloud/alien4cloud-extended-types/blob/develop/alien-extended-storage-types/alien-extended-storage-types.yml

In a first attempt, if we keep the application template default value for the Block Storage device: `/dev/vdb`, the deployment on Google Cloud fails with this error in the orchestrator:
``` 
​* google_compute_attached_disk.yorclo-blockstorage-0-attached-to-computecsdc1-0: googleapi: Error 400: Invalid value for field 'disk.deviceName': '/dev/vdb'. Must be a match of regex '(^$|\w[\w.-]{0,254})', invalid
* google_compute_attached_disk.yorclo-blockstorage-1-attached-to-computecsdc1-1: 1 error(s) occurred:
```
Maybe we could add a check on device name value on the plugin side to avoid failing later in the deployment aftar having already created compute instances ?

In a second attempt, replacing in the topology the BlockStorage property device value `/dev/vdb` by `vdb-0` in the topology,
the Block storage is created, however the component LinuxFileSystem will fail with this error :
``` 
lsblk: vdb-01: not a block device
mke2fs 1.42.9 (28-Dec-2013)
Could not stat vdb-01 --- No such file or directory
``` 

Actually, our persistent disk is available as `/dev/sdb` on the compute instance, and the device name can be found under /dev/disk/by-id/ :
```
$ ls -l /dev/disk/by-id/
lrwxrwxrwx. 1 root root  9 Oct 16 16:16 google-persistent-disk-0 -> ../../sda
lrwxrwxrwx. 1 root root 10 Oct 16 16:16 google-persistent-disk-0-part1 -> ../../sda1
lrwxrwxrwx. 1 root root  9 Oct 16 16:17 google-vdb-0 -> ../../sdb
lrwxrwxrwx. 1 root root  9 Oct 16 16:16 scsi-0Google_PersistentDisk_persistent-disk-0 -> ../../sda
lrwxrwxrwx. 1 root root 10 Oct 16 16:16 scsi-0Google_PersistentDisk_persistent-disk-0-part1 -> ../../sda1
lrwxrwxrwx. 1 root root  9 Oct 16 16:17 scsi-0Google_PersistentDisk_vdb-0 -> ../../sdb
```

so this implementation of the abstract tosca.nodes.BlockStorage is not compatible with the node alien.nodes.LinuxFileSystem  making application templates relying on this node type not portable across infrastructures.

Not sure how to deal with this issue, what do you think ?

loicalbertin:
> Could check a persistent disk was attached to the Compute instance,
> however the Application template ConsulOnBS at https://github.com/ystia/forge/tree/develop/org/ystia/experimental/consul/topologies/ConsulOnBS can't be deployed on Google Cloud due to a difference of meaning for the device name on yorc.nodes.google.PersistentDisk, compared to what expect the Alien4Cloud types defined at https://github.com/alien4cloud/alien4cloud-extended-types/blob/develop/alien-extended-storage-types/alien-extended-storage-types.yml
> 
> In a first attempt, if we keep the application template default value for the Block Storage device: `/dev/vdb`, the deployment on Google Cloud fails with this error in the orchestrator:
> 
> ```
> ​* google_compute_attached_disk.yorclo-blockstorage-0-attached-to-computecsdc1-0: googleapi: Error 400: Invalid value for field 'disk.deviceName': '/dev/vdb'. Must be a match of regex '(^$|\w[\w.-]{0,254})', invalid
> * google_compute_attached_disk.yorclo-blockstorage-1-attached-to-computecsdc1-1: 1 error(s) occurred:
> ```
> Maybe we could add a check on device name value on the plugin side to avoid failing later in the deployment aftar having already created compute instances ?
> 
> In a second attempt, replacing in the topology the BlockStorage property device value `/dev/vdb` by `vdb-0` in the topology,
> the Block storage is created, however the component LinuxFileSystem will fail with this error :
> 
> ```
> lsblk: vdb-01: not a block device
> mke2fs 1.42.9 (28-Dec-2013)
> Could not stat vdb-01 --- No such file or directory
> ```
> Actually, our persistent disk is available as `/dev/sdb` on the compute instance, and the device name can be found under /dev/disk/by-id/ :
> 
> ```
> $ ls -l /dev/disk/by-id/
> lrwxrwxrwx. 1 root root  9 Oct 16 16:16 google-persistent-disk-0 -> ../../sda
> lrwxrwxrwx. 1 root root 10 Oct 16 16:16 google-persistent-disk-0-part1 -> ../../sda1
> lrwxrwxrwx. 1 root root  9 Oct 16 16:17 google-vdb-0 -> ../../sdb
> lrwxrwxrwx. 1 root root  9 Oct 16 16:16 scsi-0Google_PersistentDisk_persistent-disk-0 -> ../../sda
> lrwxrwxrwx. 1 root root 10 Oct 16 16:16 scsi-0Google_PersistentDisk_persistent-disk-0-part1 -> ../../sda1
> lrwxrwxrwx. 1 root root  9 Oct 16 16:17 scsi-0Google_PersistentDisk_vdb-0 -> ../../sdb
> ```
> so this implementation of the abstract tosca.nodes.BlockStorage is not compatible with the node alien.nodes.LinuxFileSystem making application templates relying on this node type not portable across infrastructures.
> 
> Not sure how to deal with this issue, what do you think ?

1. Portability is the key. We should deal with with it.
2. What is the attribute value returned by Terraform on the `google_compute_attached_disk` for this device?
3. What really counts is to set an attribute that allow the linux file system to retrieve the actual device path in order to mount & format it properly. After all this property is purely indicative on OpenStack for instance if you specify `/dev/vdb` and this one already exists then `/dev/vdc` will automatically be chosen. 

So what I propose is to check if the given device is a path or a gcp compatible name, if it's a path then we transform it into a gcp compatible device name. In any case we retrieve the actual device __path__ and we set it as an attribute.

WDYT?

stebenoist:
Terraform doesn't return any attribute with the logical device name.
So e need to retrieve it in /dev/disk/by-id/google-* path after setting a specific Google device name.
I don't feel comfortable with the device property which could be used to set both logical device name or "Google reflecting device name". 
Why don't we propose another prop for Google device name ?

stebenoist:
The issue below is fixed. We got a mounted linux partition with LinuxFileSystem Alien4cloud node.
I've also fixed an issue on install workflow BlockStorage/Compute swap.
Still remains the same problem on uninstall wf but I don't have enough time to fix it right now.

PR #193, opened 2018-11-08
Install pyopenssl lib by default in the Docker iamge

adidanes:
# Pull Request description

Install pyopenssl python lib in order to be able to use Ansible modules that use openssl to generate certificates


laurentganne:
Doc change needed too in doc/install.rst (similarly to the documentation of jmespath)

loicalbertin:
I think we should take the opportunity of this PR to use specific versions of those pip dependencies in order to have more reproductible builds. If we do so documentation should be updated accordingly.
WDYT?

adidanes:
Maybe we'll re-open latter to consider Loïc suggestions

Issue #194, opened 2018-11-12
Upgrade Ansible to version 2.7


adidanes:
**As:** Yorc team
**I want to:** upgrade to 2.7 anible version
**So that:** I can benefit from updates in the crypto modules

**AC1:** Be able to generate certificates using an own CA

loicalbertin:
We should check for deprecations

adidanes:
When testing deployment of Consul with Yorc, I identified 2 warning types:

1. Using tests as filters is deprecated. Instead of using `result|changed` use `result is changed`. Remove in **2.9**
2. Invoking "yum" or "apt" (or ...) only once while using a loop via squash_actions is deprecated. Instead of using a loop to supply multiple items and specifying `name: {{item}}`, please use `name: [u'zip', u'unzip']` and remove the loop. Remove in **2.11**

PR #204, opened 2018-11-21
Ansible version updated from 2.6.3 to 2.7.2

adidanes:
# Pull Request description

Upgrade Ansible to version 2.7 in order to benefit from updates in the crypto modules

## Applicable Issues

 #194


loicalbertin:
Forge types https://github.com/ystia/forge/blob/develop/org/ystia/ansible/pub/types.yaml#L36 should also be updated

Issue #212, opened 2018-12-04
ConnectTo relationship not working for kubernetes component


Loic-R:
When deploying a topolgy for kubernetes connecting two deployments (Grafana for example), get attribute from target is working for port but for IP, it results in a "${SERVICE_IP_LOOKUP0}" which is not working for the pod

PR #213, opened 2018-12-04
Bugfix for GH-212 - SERVICE_IP_LOOKUP not working

Loic-R:
# Pull Request description

## Description of the change
Call to function replaceServiceIPInDeploymentSpec was not saved in variable deploymentRepr so not given to k8s API

### What I did
I added an unmarshaling call into the create operation (maybe it could be done in a better way)

### How to verify it
Deploy a kubernetes component where 2 deployments are connected (for example grafana in yorc-a4c-plugin/tosca-sample) 

Check yaml of grafana deployment that env variable 'GRAPHITE_URL' is correctly set


adidanes:
Tested with monitoring sample based on grafana and graphite

Issue #215, opened 2018-12-05
IP missing for k8s_service_url in runtime view - GKE


Loic-R:
# Issue Type

Put an X between the brackets of the corresponding issue type

* [X] Bug report
* [ ] Documentation issue report
* [ ] Improvement request

## Description
When deploying an app with alien, on GKE infra ; in runtime view, IP is missing for "k8s_service_url" attribute. 

## Expected behavior
k8s_service_url should display "http://IP:PORT"

## Actual behavior
Actually when apps are deployed it displays "http://:PORT"

## Steps to reproduce the issue

1. Have a correctly configured yorc and alien with kubernetes on GKE location
2. Deploy any app associated with a service (simple appache for example)

## Additional information you deem important (e.g. issue happens only occasionally)
Seems coming from : https://github.com/ystia/yorc/blob/develop/prov/kubernetes/execution.go#L362 

## Additional environment details (infrastructure implied: AWS, OpenStack, etc.)
Infra : Kubernetes on GKE

loicalbertin:
Closed per #219

PR #219, opened 2018-12-06
Bugfix/gh 215 k8s service ip missing runtime view

Loic-R:
# Pull Request description

## Description of the change
Make ip visible in runtime view of alien4cloud for services deployed on a kubernetes hosted by GKE

### What I did
I made call to kubernetes API to get external IP of nodes

### How I did it
I just added 2 functions that do the job in "utils.go" file

### How to verify it
Deploy any app on GKE that have a service, for example simple appache, and check in alien runtime view that IP is present for attribute "k8s_service_url" 

## Applicable Issues
Bug https://github.com/ystia/yorc/issues/215


prov/kubernetes/k8s_utils.go
```diff
@@ -241,3 +241,34 @@ func getNamespace(deploymentID string, objectMeta metav1.ObjectMeta) (string, bo
...
+		return "", errors.Wrap(err, "Failed to get nodes")
+	}
+	for _, node := range nodes.Items {
+		for _, cond := range node.Status.Conditions {
+			if cond.Type == corev1.NodeReady && cond.Status == corev1.ConditionTrue {
+				return node.ObjectMeta.Name, nil
+			}
+		}
+	}
+	//No healthy node found (error or warning to throw maybe)
```
loicalbertin: Indeed at least a warning should be printed.
So I suggest to return an error here and to emit a log event on use of getHealthyNode if no node could be retrieved.
We should not stop IMHO but simply generate an empty attribute or better just the port part.

prov/kubernetes/execution.go
```diff
@@ -520,11 +516,9 @@ func (e *execution) deployNode(ctx context.Context, clientset kubernetes.Interfa
 		}
 		var s string
 		for _, val := range serv.Spec.Ports {
-			kubConf := e.cfg.Infrastructures["kubernetes"]
-			kubMasterIP := kubConf.GetString("master_url")
-			u, _ := url.Parse(kubMasterIP)
-			h := strings.Split(u.Host, ":")
-			str := fmt.Sprintf("http://%s:%d", h[0], val.NodePort)
+			node, _ := getHealthyNode(clientset)
```
loicalbertin: why is this done in the for loop? while in the previous use above it is done outside of the loop?
This should reduce API usage.

prov/kubernetes/k8s_utils.go
```diff
@@ -241,3 +241,34 @@ func getNamespace(deploymentID string, objectMeta metav1.ObjectMeta) (string, bo
...
+	node, err := clientset.CoreV1().Nodes().Get(nodeName, metav1.GetOptions{})
+	if err != nil {
+		return "", errors.Wrap(err, "Failed to get node "+nodeName)
+	}
+	for _, addr := range node.Status.Addresses {
+		if addr.Type == corev1.NodeExternalIP {
+			return addr.Address, nil
+		}
+	}
+	return "", errors.Wrap(err, "Node "+nodeName+" don't have external IP adress")
```
loicalbertin: Is this possible?
What if we are not using GKE but another cluster?

prov/kubernetes/execution.go
```diff
@@ -355,14 +354,11 @@ func (e *execution) manageServiceResource(ctx context.Context, clientset kuberne
 		}
 
 		events.WithContextOptionalFields(ctx).NewLogEntry(events.LogLevelDEBUG, e.deploymentID).Registerf("k8s Service %s created in namespace %s", service.Name, namespace)
-
-		kubConf := e.cfg.Infrastructures["kubernetes"]
-		kubMasterIP := kubConf.GetString("master_url")
-		u, _ := url.Parse(kubMasterIP)
-		h := strings.Split(u.Host, ":")
+		node, _ := getHealthyNode(clientset)
```
loicalbertin: Ignoring errors is a bad practice...
You should at least log it (better is to emit an event log that would be seen by the user).
Errcheck + log  does not prevent to continue anyway with an empty string.

prov/kubernetes/k8s_utils.go
Loic-R: I don't really know... Maybe as you said, if it is not a GKE cluster and cluster admin decided to make his worker nodes sort of 'private'

Issue #214, opened 2018-12-04
No output properties for services on GKE


Loic-R:
# Issue Type

Put an X between the brackets of the corresponding issue type

* [X] Bug report
* [ ] Documentation issue report
* [ ] Improvement request

## Expected behavior
See ip and node port of the service in alien output properties

## Actual behavior
Output properties diplays : "N/A"

## Steps to reproduce the issue

1. Configure a yorc/alien with a kubernetes infrastructure
2. Deploy any kubernetes topology that have an associated service

## Additional information you deem important (e.g. issue happens only occasionally)
It seems that it comes from this line : https://github.com/ystia/yorc/blob/develop/prov/kubernetes/execution.go#L362

## Additional environment details (infrastructure implied: AWS, OpenStack, etc.)
Infra : Kubernetes on google cloud engine

loicalbertin:
Resolved per #223

PR #223, opened 2018-12-10
Bugfix/gh 214 no ouptut properties gke services

Loic-R:
# Pull Request description

## Description of the change
Make output properties of services visible alien4cloud for services deployed on a kubernetes hosted by GKE

### What I did
Set "node_port" attribute that was not set before

### How I did it
I just make a call of function setAttributeForAllInstances

### How to verify it
Deploy any app on GKE that have a service, for example simple appache, and check in alien "info" of "manage current deployment" tab that "node_port" output propertie is set and is no longer "N/A"

## Applicable Issues
Bug https://github.com/ystia/yorc/issues/214

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## UNRELEASED
 
+### BUG FIXES
+
+* No output properties for services on GKE ([GH-214](https://github.com/ystia/yorc/issues/214))
+
 ## 3.1.0-M7 (December 07, 2018)
 
 ### DEPENDENCIES
--- a/prov/kubernetes/execution.go
+++ b/prov/kubernetes/execution.go
@@ -368,6 +368,10 @@ func (e *execution) manageServiceResource(ctx context.Context, clientset kuberne
 				if err != nil {
 					return errors.Wrap(err, "Failed to set attribute")
 				}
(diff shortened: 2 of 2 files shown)


adidanes:
> @adidanes I think you are talking about another bug opened, [pull request](https://github.com/ystia/yorc/pull/219) is already done for it but maybe need some changes

Indeed !

loicalbertin:
> > @adidanes I think you are talking about another bug opened, [pull request](https://github.com/ystia/yorc/pull/219) is already done for it but maybe need some changes
> 
> Indeed !

As long as #219 is not merged we will have this issue.

adidanes:
I will test the #219

adidanes:
I would like to test both issues once. 
Is it possible to merge one of the branches into the other one ?

adidanes:
Maybe it would be usefull to have an attribute set to the service's endpoint.
Ex : 10.23.250.151:80 TCP

Loic-R:
> Maybe it would be usefull to have an attribute set to the service's endpoint.
> Ex : 10.23.250.151:80 TCP

Do you mean the cluster internal service endpoint ? Not sure if it is useful but we can set an attribute yes.

> However, the k8s_service_url can't be checked (NodePort not opened ?).

Yes I don't know how why but even if firewall rules are created on GCP, we can't access public IP adress of our nodes.. 

> Moreover, I note that the <NODE_IP> does not correspond to the one of the Managed Pod
Actually, it takes the first external IP found but that's not a problem if the pod is on another node because request are redirected on it

PR #224, opened 2018-12-10
Fix issue on cuda_visible_devices command and output

### LINKED ISSUE #210: CUDA_VISIBLE_DEVICES contains some unwanted unprintable characters
# Issue Type

* [X] Bug report
* [ ] Improvement request

## Description
Hey !

When trying to deploy a Slurm compute using gres, it feels like the returned CUDA_VISIBLE_DEVICES environment variable contains some unwanted unprintable caracters.
Trying to use it with nvidia-docker 2, I got the following issue : 

/usr/bin/nvidia-container-cli --load-kmods configure --ldconfig=@/sbin/ldconfig.real --device=0\\\\r --compute --utility --require=cuda>=9.0 --pid=30288 /var/lib/docker/overlay2/0bb0b79b9132e98e42186cccbcfdf64a99631c3baed67f5af253bd775da1f228/merged]\\\\nnvidia-container-cli: device error: unknown device id: 0\\\\r\\\\n\\\

I resolved such issue while doing inside my scripts : 
echo ${CUDA_VISIBLE_DECIVES}| tr -d '\1-\31 '

I think that would be better to be treated on Yorc side.
## Expected behavior
CUDA_VISIBLE_DEVICES variable is returned without unwanted characters

1. Step1
2. Step2 ...

## Output of `yorc version`
3.0.0
## Priority

Medium

stebenoist:
## Description of the change

Change command to retrieve CUDA_VISIBLE_DEVICES
Trim "/r/n" in output

## Applicable Issues
fixes #210


prov/slurm/helper.go
```diff
@@ -112,11 +112,12 @@ func getAttributes(client sshutil.Client, key string, params ...string) ([]strin
 	switch key {
 	case "cuda_visible_devices":
 		if len(params) == 2 && params[0] != "" {
-			cmd := fmt.Sprintf("srun --jobid=%s env|grep CUDA_VISIBLE_DEVICES", params[0])
+			cmd := fmt.Sprintf("srun --jobid=%s  bash -c 'env|grep CUDA_VISIBLE_DEVICES'", params[0])
 			stdout, err := client.RunCommand(cmd)
 			if err != nil {
 				return nil, errors.Wrapf(err, "Unable to retrieve (%s) for node:%q", key, params[1])
 			}
+			stdout = strings.Trim(stdout, "\r\n")
```
loicalbertin: No more \x00 to trim?
Like bellow for the partition

stebenoist: Done before here: https://github.com/ystia/yorc/blob/bugfix/gh-210-cuda-visible-dev/helper/sshutil/sshutil.go#L103

loicalbertin: ok good

PR #246, opened 2018-12-19
Increase default workers number per Yorc server from 3 to 30

### LINKED ISSUE #244: Increase default workers number
# Issue Description

As per our recent changes on making workflow steps the new unit of execution instead of the whole workflow it makes sense to increase the default number of workers from `3` to lets say `30`. This will be adapted when we will have real use-cases feedback

loicalbertin:
# Pull Request description

## Description of the change

Increase default workers number per Yorc server from `3` to `30` 

* Increase default workers number per Yorc server from `3` to `30` ([GH-244](https://github.com/ystia/yorc/issues/244))

## Applicable Issues

Resolves #244

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## UNRELEASED
 
+### ENHANCEMENTS
+
+* Increase default workers number per Yorc server from `3` to `30` ([GH-244](https://github.com/ystia/yorc/issues/244))
+
 ## 3.1.0-RC2 (December 18, 2018)
 
 ### DEPENDENCIES
--- a/config/config.go
+++ b/config/config.go
@@ -33,7 +33,7 @@ import (
 const DefaultConsulPubMaxRoutines int = 500
 
 // DefaultWorkersNumber is the default number of workers in the Yorc server
-const DefaultWorkersNumber int = 3
+const DefaultWorkersNumber int = 30
 
 // DefaultHTTPPort is the default port number for the HTTP REST API
 const DefaultHTTPPort int = 8800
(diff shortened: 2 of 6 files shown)


loicalbertin:
> How the number 30 was chosen ?
> 
> Otherwise, looks good

It should be a trade-off between how many steps we want to allow to run in parallel which impacts overall Yorc's performances and the number of terraform/ansible executions run in parallel by Yorc that will impact consumed system memory and CPU.

So the value of `30` sounds reasonable for standard cloud servers that generally hosts Yorc. 
This is still configurable for constrained or large hosts.

Loic-R:
> > How the number 30 was chosen ?
> > Otherwise, looks good
> 
> It should be a trade-off between how many steps we want to allow to run in parallel which impacts overall Yorc's performances and the number of terraform/ansible executions run in parallel by Yorc that will impact consumed system memory and CPU.
> 
> So the value of `30` sounds reasonable for standard cloud servers that generally hosts Yorc.
> This is still configurable for constrained or large hosts.

Ok I see. Maybe we should test if it doesn't cause performance issues when it is running in a containerized environment.

loicalbertin:
> > > How the number 30 was chosen ?
> > > Otherwise, looks good
> > 
> > 
> > It should be a trade-off between how many steps we want to allow to run in parallel which impacts overall Yorc's performances and the number of terraform/ansible executions run in parallel by Yorc that will impact consumed system memory and CPU.
> > So the value of `30` sounds reasonable for standard cloud servers that generally hosts Yorc.
> > This is still configurable for constrained or large hosts.
> 
> Ok I see. Maybe we should test if it doesn't cause performance issues when it is running in a containerized environment.

We already had feedback on the previous model (1 worker per workflow) and up to 50 workers both on containerized and non-containerized environments. Results were good. So this modification seems a bit conservative to me. But we will see later on if we need increase again this default.

Issue #245, opened 2018-12-19
K8S jobs namespace should not be removed if its provided


adidanes:
# Issue Type

Put an X between the brackets of the corresponding issue type

* [x] Bug report
* [ ] Documentation issue report
* [ ] Improvement request

## Description
User can provide a namespace name for the application to be deployed. When the application contains K8S Jobs, the Jobs should run in the context of the provided namespace, and when the job ends, the namespace still remains.

## Expected behavior
When the job ennds, the namespace should remain.

## Actual behavior
The namespace is deleted when the job ends.

## Steps to reproduce the issue

1. Step1
2. Step2 ...

## Output of `yorc version`
V3.1.0-RC2

## Priority

Low

* High = We should stop anything we are doing and do this. If you're writing this calmly, you should probably choose another option. If it's really that urgent, please make sure we read this.
* Medium = It's important that we do this within a few days.
* Low = We will consider this on our next sprint planning.

(Please be aware that your priority may not match ours, we'll use this as guidance only).

loicalbertin:
Should be backported to 3.1.1 release.

PR #262, opened 2019-01-03
Bugfix/GH 245 - K8S jobs namespace should not be removed if its provided 

Loic-R:
# Pull Request description

### What I did
Make yorc not remove namespace when it is provided

### How I did it
I change a wrong condition and key for a map

### How to verify it
Deploy 2 K8s job apps : 

1. one  setting the meta property "K8S_NAMESPACE" to an existing namespace on the cluster (default for example)
2. the other app not setting this meta property

Check that in case 1, it does not remove or try to remove the namespace where the job is executed ; and in case 2 that it delete the namespace once the job is finished.

## Applicable Issues
https://github.com/ystia/yorc/issues/245

### DIFF
--- a/prov/kubernetes/monitoring_job.go
+++ b/prov/kubernetes/monitoring_job.go
@@ -53,8 +53,8 @@ func (o *actionOperator) ExecAction(ctx context.Context, cfg config.Configuratio
 	}
 	// Check if namespace was provided
 	var namespaceProvided bool
-	namespaceProvidedStr, ok := action.Data["providedNamespace"]
-	if b, err := strconv.ParseBool(namespaceProvidedStr); ok && err != nil {
+	namespaceProvidedStr, ok := action.Data["namespaceProvided"]
+	if b, err := strconv.ParseBool(namespaceProvidedStr); ok && err == nil {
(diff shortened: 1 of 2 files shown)


prov/kubernetes/monitoring_job.go
```diff
@@ -53,8 +53,8 @@ func (o *actionOperator) ExecAction(ctx context.Context, cfg config.Configuratio
 	}
 	// Check if namespace was provided
 	var namespaceProvided bool
-	namespaceProvidedStr, ok := action.Data["providedNamespace"]
-	if b, err := strconv.ParseBool(namespaceProvidedStr); ok && err != nil {
+	namespaceProvidedStr, ok := action.Data["namespaceProvided"]
+	if b, err := strconv.ParseBool(namespaceProvidedStr); ok && err == nil {
```
stebenoist: Maybe we should check first ok is true before converting namespaceProvidedStr to bool ?

Loic-R: Yes in fact. Change done

adidanes:
Tested. Works good for me.

Issue #260, opened 2018-12-21
Factorize and make documentation more user friendly


loicalbertin:
**As:** your favorite user
**I want to:** have one single pointer to access documentation
**So that:** I can find anything I want in 2 seconds

* [ ] **AC1:** All documentation accessible within one single location
* [ ] **AC2:** Should be able to version some way documentation

PR #270, opened 2019-01-10
Move Bintray for releases and Artifactory for dev artifacts

loicalbertin:
Move releases to BinTray and development artifacts to Artifactory


build/deploy_artifactory.sh
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env bash
+
+scriptDir="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+rootDir=$(readlink -f "${scriptDir}/..")
```
stebenoist: rootDir appears unused. (Cf CodeClimate)

loicalbertin: It is used now.

good point for the documentation you are pointing out, but it is global to all Ystia products so maybe its migration to a public place should be addressed in #260

Issue #274, opened 2019-01-15
Can't deploy applications using a secured yorc/consul


laurentganne:
# Issue Type

* [X] Bug report
* [ ] Documentation issue report
* [ ] Improvement request

## Description

Attempting to deploy an application on a secured setup (both Yorc and consul secured), the deployment fails early in the preliminary steps when Yorc is storing key/values pairs in Consul:
at some point, after having stored many key/values already, this TLS handshake error appears
```
[DEBUG] ERROR: Put https://127.0.0.1:8543/v1/kv/_yorc/deployments/test/topology/types/tosca.nodes.WebServer/capabilities/admin_endpoint/occurrences/lower_bound?dc=dc1: net/http: TLS handshake timeout
Failed to store consul key "_yorc/deployments/test/topology/types/tosca.nodes.WebServer/capabilities/admin_endpoint/occurrences/lower_bound"
github.com/ystia/yorc/helper/consulutil.(*rateLimitedConsulPublisher).publish.func1
/home/travis/gopath/src/github.com/ystia/yorc/helper/consulutil/consul_ratelimit_publisher.go:162
github.com/ystia/yorc/vendor/golang.org/x/sync/errgroup.(*Group).Go.func1
/home/travis/gopath/src/github.com/ystia/yorc/vendor/golang.org/x/sync/errgroup/errgroup.go:58
runtime.goexit
```
And the deployment fails.

Also, another error is that when Yorc configures Terraform to store its state in Consul, Yorc doesn't specify a certificate authority, certificate and key.
So, even if the handshake error described above was fixed, the deployment would fail later during the on-demand compute instance creation with this error :
```
Error refreshing state: Get https://127.0.0.1:8543/v1/kv/_yorc/deployments/test/terraform-state/Compute: x509: certificate signed by unknown authority

Status for workflow step "Compute_install" changed to "ERROR"
...
Failed to setup Consul remote backend for terraform
github.com/ystia/yorc/prov/terraform.(*defaultExecutor).remoteConfigInfrastructure
        /home/vscode/go/src/github.com/ystia/yorc/prov/terraform/executor.go:180
github.com/ystia/yorc/prov/terraform.(*defaultExecutor).applyInfrastructure
        /home/vscode/go/src/github.com/ystia/yorc/prov/terraform/executor.go:255
github.com/ystia/yorc/prov/terraform.(*defaultExecutor).installNode
        /home/vscode/go/src/github.com/ystia/yorc/prov/terraform/executor.go:107
[...]

PR #275, opened 2019-01-16
TLS handhake timeout during deployment using yorc with a secured consul

laurentganne:
# Pull Request description

## Description of the change

Fixed issue https://github.com/ystia/yorc/issues/274 :  Can't deploy applications using a secured yorc/consul

### What I did

Added missing settings in prov/terraform/*/generator.go for Terraform to use a secured consul as backend.

Added a new Yorc Server Consul config parameter to specify a maximum duration to wait for a TLS handshake with Consul, with a default value of 50 seconds.
This is needed as Yorc is making an intensive use of Consul during the deployments preliminary phase when Yorc is storing deployment details in Consul.
The CPU consumption is high in this phase, having the side effect of increasing the TLS handshake duration with Consul, so much that the default TLS handshake timeout of 10 seconds is too low, as already described at https://github.com/hashicorp/consul/issues/3711.
The deployment performed by Yorc will then fail on TLS handshake timeout if the default timeout is used, during this preliminary step where Yorc is storing deployment details in Consul.

Can't deploy applications using a secured yorc/consul ([GH-274](https://github.com/ystia/yorc/issues/274))

## Applicable Issues

https://github.com/ystia/yorc/issues/274

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -98,6 +98,10 @@ Globals Command-line options
 
   * ``--consul_ssl_verify``: If set to false, disable Consul certificate checking (true by default is ssl enabled).
 
+.. _option_consul_tls_handshake_timeout_cmd:
+
+  * ``--consul_tls_handshake_timeout``: Maximum duration to wait for a TLS handshake with Consul, the default is ``50s``.
+
 .. _option_terraform_plugins_dir_cmd:
 
   * ``--terraform_plugins_dir``: Specify the directory where to find Terraform pre-installed providers plugins. If not specified, required plugins will be downloaded during deployment. See https://www.terraform.io/guides/running-terraform-in-automation.html#pre-installed-plugins for more information.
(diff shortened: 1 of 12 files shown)


doc/configuration.rst
```diff
@@ -723,6 +731,11 @@ Environment variables
 
   * ``YORC_CONSUL_SSL_VERIFY``: Equivalent to :ref:`--consul_ssl_verify <option_consul_ssl_verify_cmd>` command-line flag.
 
+.. _option_consul_tls_handshake_timeout_env:
+
+  * ``YORC_CONSUL_TLS_HANDSHAKE_TIMEOUT``: Equivalent to :ref:`--consul_tls_handshake_timeout <option_consul_tls_handshake_timeout_cmd>` command-line flag.
```
loicalbertin: no way/need to give this timeout to terraform?

prov/terraform/openstack/generator.go
```diff
@@ -94,9 +94,12 @@ func (g *osGenerator) GenerateTerraformInfraForNode(ctx context.Context, cfg con
 	infrastructure.Terraform = map[string]interface{}{
 		"backend": map[string]interface{}{
 			"consul": map[string]interface{}{
-				"path":    terraformStateKey,
-				"address": consulAddress,
-				"scheme":  consulScheme,
+				"path":      terraformStateKey,
```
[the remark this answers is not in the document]
laurentganne: ok, refactored in last push

doc/configuration.rst
laurentganne: Didn't find such config param and didn't need it in my test as the test was basic, just doing one deployment at a time. in this case, it was needed only at the preliminary phase of the deployment storing the deployment details in Consul, when Terraform is not yet involved.

loicalbertin: ok

PR #301, opened 2019-02-04
Add events on workflow change for purge task

stebenoist:
# Pull Request description

Missing status events for workflow of task purge typed


stebenoist:
Hum I need to create a new issue on yorc side instead of referring to the yorc-a4c-plugin one

PR #339, opened 2019-03-15
Bugfix/gh 324 add ctx props to events

### LINKED ISSUE #324: Custom command events doesn't provide enough information
# Issue Type

* [X] Bug report
* [ ] Improvement request

## Description
Custom command events doesn't provide any information about nodeID, operationName.
However, this information is present in the context.
Overall, I wonder why we don't retrieve information from context in all event publications.

## Priority

Medium

stebenoist:
# Pull Request description

### What I did
1. Use the context to build event information for all events
2. Emit event on status change when related task status is updated
3. Add more information on operation and node name in case of custom command
4. add this information in events CLI

### How to verify it
Deploy the python component and execute custom commands say_hello/say_goofbye
(https://github.com/ystia/yorc-a4c-plugin/tree/develop/tosca-samples/org/ystia/yorc/samples/python)
Check with CLI (yorc d events -b) that operation name and node are present in the custom command related events

* Custom command events doesn't provide enough information ([GH-324](https://github.com/ystia/yorc/issues/324))
## Applicable Issues
fixes #324


prov/ansible/execution.go
```diff
@@ -61,6 +61,7 @@ host_key_checking=False
 timeout=30
 stdout_callback = yaml
 retry_files_save_path = #PLAY_PATH#
+nocows = 1
```
loicalbertin: Hum looks like very depend to your env :laughing:

tasks/workflow/task_execution.go
```diff
@@ -180,8 +181,18 @@ func setTaskStatus(kv *api.KV, taskID string, status tasks.TaskStatus, lastIndex
 	}
 	if !set {
 		log.Debugf("[WARNING] Failed to set task status to:%q for taskID:%q as last index has been changed before. Retry it", status.String(), taskID)
-		return checkAndSetTaskStatus(kv, taskID, status)
+		return checkAndSetTaskStatus(ctx, kv, targetID, taskID, status)
 	}
+
+	// Emit event for status change
+	// As task has been set, error are ignored
+	wfName, _ := tasks.GetTaskData(kv, taskID, "workflowName")
```
loicalbertin: What will happen if wfName is empty (in case of error silently bypassed)?

prov/ansible/execution.go
stebenoist: Yep, but we don't support cows in Ansible stdout handler. And in some distribs, I'm afraid it's the default behavior ?
So let's disable it.

tasks/workflow/task_execution.go
stebenoist: wfName may be empty as this data is not filled for non-workflow task type (as for custom command by instance)

stebenoist: I add some comments

prov/ansible/execution.go
loicalbertin: That's a pity but ok :smiley:

tasks/workflow/task_execution.go
loicalbertin: Sounds good.

loicalbertin:
Code review : LGTM

Issue #351, opened 2019-03-26
Better detection and feedback on Slurm & K8S jobs statuses


loicalbertin:
**As:** An app designer
**I want to:** have a better feedback on actual status of Slurm & K8S Jobs
**So that:** I can see if they are running or queued

**AC1:** Emit events and change status of jobs to reflect their actual status
**AC2:** in the run operation emit non-standard statuses
**AC3:** for Slurm and kubernetes

PR #363, opened 2019-04-12
Feature/gh 351 jobs status feedback

adidanes:
# Pull Request description

Need to provide more precise information about the jobs state. Focus on slurm and k8s jobs.

## Description of the change

Get specific status information on slurm/k8s jobs. 
Set this information in the job instances 'state' attribute. 
Emit events and logs corresponding to this state attribute updates.

### What I did

Updated the monitoring function of the slurm/k8s provider. This function is periodically executed to check status infos about the job, trace status info, and thus allow to detect job execution end.

### How I did it

Get specific slurm/k8s status information about running jobs. Use the 

`SetInstanceStateStringWithContextualLogs `

function from the deployments `package `in order to :

- update **state** attribute
- generate **state** attribute update event
- generate log about **state** attribute update

### How to verify it

Use Job samples from alien4cloud-yorc-plugin project

#### Test on K8S (GKE)
Be ready to resize target K8S cluster (gcloud container cluster resize) to add/remove resources

Create, deploy application containing a job (KubeJobPi, etc)
Execute run wf
Check logs and events (.... ToBe detailed ...)
Check state attribute in consul 
$ docker exec -it klt-yorc-demo-lwyn bash
$ consul kv get -recurse _yorc/deployments/jobPi | grep "topology/instances" | grep state

## Applicable Issues
https://github.com/ystia/yorc/issues/351


prov/slurm/monitoring_jobs.go
```diff
@@ -134,7 +139,6 @@ func (o *actionOperator) monitorJob(ctx context.Context, cfg config.Configuratio
 		// job's still running or its state is about to be set definitively: monitoring is keeping on
 	default:
 		// Other cases as FAILED, CANCELLED, STOPPED, SUSPENDED, TIMEOUT, etc : error is return with job state and job info is logged
-		events.WithContextOptionalFields(ctx).NewLogEntry(events.LogLevelERROR, deploymentID).RegisterAsString(fmt.Sprintf("job info:%+v", info))
```
stebenoist: This log event is useful as it displays the whole Slurm information in case of failure

Loic-R:
Tested on GKE. 
Works well with KubeJobPi but with KubeJobFail, it has a strange comportment : 
After job failed, instance changes to an empty status and keep returning empty status until job retry, it changes to "Running" and after the fail ; again to an empty status.

After 15min, status changed to "Failed" so that part seems working well :heavy_check_mark: 

For more info see logs below :
[fail-Environment-0.1.0-SNAPSHOT.log](https://github.com/ystia/yorc/files/3089525/fail-Environment-0.1.0-SNAPSHOT.log)

adidanes:
Question to @Loic-R : when you say that **status** is emty, you make reference to **state** attribute of the job instance ?

Loic-R:
> Question to @Loic-R : when you say that **status** is emty, you make reference to **state** attribute of the job instance ?

Yes in fact it displays an empty string like this : 
[run][]Status for node "ContainerJobUnit_Resource", instance "0" changed to ""

prov/slurm/monitoring_jobs.go
```diff
@@ -125,6 +128,8 @@ func (o *actionOperator) monitorJob(ctx context.Context, cfg config.Configuratio
 		o.logFile(ctx, deploymentID, fmt.Sprintf("slurm-%s.out", actionData.jobID), "StdOut/Stderr", sshClient)
 	}
 
+	deployments.SetInstanceStateStringWithContextualLogs(ctx, consulutil.GetKV(), deploymentID, action.Data["nodeName"], "0", info["JobState"])
```
stebenoist: We should only update the status to "jobState" if not already set to this value

adidanes:
Good catch @Loic-R , the empty status seems to be due to situations when no pods are running, and no pods in either failed or succeeded state.

adidanes:
Thanks teammates !

Issue #368, opened 2019-04-18
Unable to delete a deployment with non-conform topology


stebenoist:
# Bug Report

## Description

Unable to delete a deployment with non-conform topology

## Expected behavior
When a deployment is provided with non conform topology, the deployment status is failed and it's impossible to undeploy and purge it.

## Steps to reproduce the issue
Use deployments related to https://github.com/ystia/yorc/issues/367

PR #369, opened 2019-04-18
Force purge to delete deployment with inconsistent topology

stebenoist:
# Pull Request description

## Description of the change
When a deployment with non-conform topology is submitted to yorc, no topology template is stored.
When we try to undeploy it, no task execution is created as no workflow steps has been found. Only the main task is created and so nothing is done.

### What has been done

An "inconsistentDeploymentError" error is returned by task collector when no workflows step is detected.
A "Force purge task" is ordered to allow removing the deployment.

### How to verify it
See https://github.com/ystia/yorc/issues/368
* Unable to delete a deployment with non-conform topology ([GH-368](https://github.com/ystia/yorc/issues/368))
## Applicable Issues
#368

### DIFF
--- a/tasks/workflow/builder/builder.go
+++ b/tasks/workflow/builder/builder.go
@@ -38,6 +38,10 @@ func BuildWorkFlow(kv *api.KV, deploymentID, wfName string) (map[string]*Step, e
 		return nil, err
 	}
 
+	if wf.Steps == nil || len(wf.Steps) == 0 {
+		return nil, tasks.NewInconsistentDeploymentError(deploymentID)
+	}
+
 	steps := make(map[string]*Step, len(wf.Steps))
 	visitedMap := make(map[string]*visitStep, len(wf.Steps))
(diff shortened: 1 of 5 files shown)


tasks/workflow/builder/builder.go
```diff
@@ -38,6 +38,10 @@ func BuildWorkFlow(kv *api.KV, deploymentID, wfName string) (map[string]*Step, e
 		return nil, err
 	}
 
+	if wf.Steps == nil || len(wf.Steps) == 0 {
+		return nil, tasks.NewInconsistentDeploymentError(deploymentID)
```
loicalbertin: Why does this lives in `tasks` packages and not in `deployments` package?

stebenoist: Honestly, I didn't know where to put this !

stebenoist: Done

PR #389, opened 2019-05-03
Add ticker to handle node state for monitoring

### LINKED ISSUE #388: Monitoring causes Yorc and Consul CPU over consumption
# Bug Report

## Description

When using HTTP or TCP monitoring, the node state change polling for each monitored compute increases notably both Yorc and Consul CPU usage

## Priority

High / Medium / Low

* High

stebenoist:
# Pull Request description

## Description of the change
Add a ticker to handle node state for monitoring with time interval = check time interval / 2

## How to verify it
Check CPU usage

* Monitoring causes Yorc and Consul CPU over consumption ([GH-388](https://github.com/ystia/yorc/issues/388))

## Applicable Issues

Fixes #388


prov/monitoring/check.go
```diff
@@ -110,6 +116,15 @@ func (c *Check) disable() {
 	}
 }
 
+func (c *Check) enable() {
+	if !c.enabled {
```
loicalbertin: I'm surprised to not see this under lock sync
Is it safe for concurrent use?

stebenoist: Hum, I thougt that there was no need to handle any concurrency for this one but looking closer, concurrency can happen when stopping the check

Issue #383, opened 2019-04-30
Migrate from Oracle JDK to OpenJDK when bootstrapping Yorc/Alien4Cloud


loicalbertin:
# Bug Report

## Description

There is again some issues with JDK 8 downloads from Oracle website.
As per recent announcements by RedHat on JDK support I propose to migrate to OpenJDK and download it from <https://adoptopenjdk.net> by default.

PR #392, opened 2019-05-06
Use OpenJDK instead of Oracle JDK for Yorc/Alien bootstrap

stebenoist:
# Pull Request description

## Description of the change
- Create a node type org.alien4cloud.java.jdk.linux.nodes.OpenJDK in Bootstrap Tosca types and use it for JDK template.
- Change default JDK Download URL and version

* Migrate from Oracle JDK to OpenJDK when bootstrapping Yorc/Alien4Cloud ([GH-383](https://github.com/ystia/yorc/issues/383))
## Applicable Issues
#383


commands/bootstrap/resources/topology/topology_single_node.tmpl
```diff
@@ -42,7 +42,7 @@ imports:
 topology_template:
   node_templates:
     JDK:
-      type: org.alien4cloud.java.jdk.linux.nodes.OracleJDK
+      type: org.alien4cloud.java.jdk.linux.nodes.OpenJDK
```
loicalbertin: Where is this type defined?

stebenoist: It's in the zip file tosca_types.zip

loicalbertin: Yeah but we do relie on <https://github.com/alien4cloud/csar-public-library/blob/45ab6701f8a5193558598f928e8cfc503115989a/org/alien4cloud/java/jdk/linux/types.yml#L15> I think it is a bad idea to rename it.
Furthermore we should port this change to the public csar lib too.

commands/bootstrap/inputs.go
```diff
@@ -170,11 +170,11 @@ var (
 	jdkDefaultInputs = map[string]defaultInputType{
 		"jdk.download_url": defaultInputType{
 			description: "Java Development Kit download URL",
-			value:       "https://download.oracle.com/otn-pub/java/jdk/8u202-b08/1961070e4c9b4e26a04e7f5a083f551e/jdk-8u202-linux-x64.tar.gz",
+			value:       "https://api.adoptopenjdk.net/v2/binary/releases/openjdk8?openjdk_impl=hotspot&os=linux&arch=x64&release=jdk8u212-b03&type=jdk",
```
loicalbertin: This point to https://github.com/AdoptOpenJDK/openjdk8-binaries/releases/download/jdk8u212-b03/OpenJDK8U-jdk_x64_linux_hotspot_8u212b03.tar.gz
I think I prefer this url  (more readable, less error-prone), WDYT?

stebenoist: Not sure if they decide to move their releases on Bintray or other download platform. Maybe their API is more stable ?

commands/bootstrap/resources/topology/topology_single_node.tmpl
stebenoist: I didn't rename it but add a new one. But OK to keep the existing one despite its name.

commands/bootstrap/inputs.go
loicalbertin: Ok, my concern was that when used this value should be properly quoted in order to prevent interpreting `&` characters by a shell.

PR #393, opened 2019-05-07
set ansible playbook environment proxy settings for no proxy

### LINKED ISSUE #384: Bootstrap step YorcOnDemandLocationResources fails on setup with http proxy defined
# Bug Report

## Description

On a setup having external access through a proxy, a bootstrap fails at step YorcOnDemandLocationResources when it attempts to connect to the Alien4Cloud UI,
as it tries to connect going through the proxy, even if environment variables NO_PROXY and no_proxy are set correctly.

## Output of `yorc version`

3.1.2

Affetcs as well 3.2.0 snapshots

## Priority

Medium

stebenoist:
remove defers from loops

# Pull Request description

## Description of the change
Empty proxy setting in playbook environment for all tasks for Alien OnDemand resources location creation script in bootstrap type as it's done for orchestrator creation : we assume that Alien is accessible without proxies.

* Bootstrap step YorcOnDemandLocationResources fails on setup with http proxy defined ([GH-384](https://github.com/ystia/yorc/issues/384))

## Applicable Issues

Fixes #384


commands/bootstrap/resources.go
```diff
@@ -86,7 +86,6 @@ func extractResources(resourcesZipFilePath, resourcesDir string) error {
 			if err != nil {
 				return err
 			}
-			defer dest.Close()
```
loicalbertin: Why did you removed the close? if io.Copy errors `dest`will not be closed...

commands/bootstrap/resources.go
```diff
@@ -163,7 +172,6 @@ func getVersionFromTOSCATypes(path string) string {
 			fmt.Println("Failed to open tosca types zip in bundled resources", err)
 			return version
 		}
-		defer src.Close()
```
loicalbertin: Same as above

commands/bootstrap/resources.go
stebenoist: I removed the defer from the loop. The close is below

stebenoist: But yes, it's a bad idea

loicalbertin: I missed the loop when reviewing...
If there is a defer in a loop that sounds like it should be done into a function (even an anonymous one)

commands/bootstrap/resources.go
```diff
@@ -176,6 +184,11 @@ func getVersionFromTOSCATypes(path string) string {
 			version = match[1]
 			return version
 		}
+		err = src.Close()
```
loicalbertin: When matching close will never be called.

stebenoist: I changed this !

PR #404, opened 2019-05-17
Emit an event when purge is actually done.

### LINKED ISSUE #402: A "purged" status is missing for deployments
# Bug Report

## Description

It is difficult for some external programs using the REST API to know when a deployment is actually purged from Yorc. This is particularly true for https://github.com/alien4cloud/alien4cloud-yorc-provider.
Emitting a `purged` event would help

## Expected behavior

Emit an event when a deployment is actually purged (when job is actually done by a worker as this is asynchronous)
This event should be persisted after the purge of the deployment so it can be easily retrieved thanks to the event API.

loicalbertin:
# Pull Request description

## Description of the change

Emit an event when purge is actually done.
Remove logs & events before deploying a new
application to remove those stale purged events

### How to verify it

Should use the corresponding plugin PR ystia/yorc-a4c-plugin#123.

Deploy an application and undeploy it using the CLI with the purge option.
Then redeploy the same application...

* Emit a persistent event on deployment purge ([GH-402](https://github.com/ystia/yorc/issues/402))

## Applicable Issues

Fixes #402

### DIFF
--- a/tasks/workflow/worker.go
+++ b/tasks/workflow/worker.go
@@ -628,11 +628,7 @@ func (w *worker) runPurge(ctx context.Context, t *taskExecution) error {
 	if err != nil {
 		return err
 	}
-	// Just Publish an event that the deployment is successfully
-	// This event will stay into consul even if the deployment is actually purged...
-	// To prevent unexpected errors
-	_, err = events.PublishAndLogDeploymentStatusChange(ctx, t.cc.KV(), t.targetID, strings.ToLower(deployments.PURGED.String()))
-	return err
+	return deployments.TagDeploymentAsPurged(ctx, t.cc, t.targetID)
 }
 
 func (w *worker) runScaleOut(ctx context.Context, t *taskExecution) error {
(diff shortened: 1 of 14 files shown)


laurentganne:
May be an issue for client applications like fastML deploying applications with timestamped application names or using UUIDs in names of application it deployes: in this case events and logs of purged applications would not be purged in consul.

tasks/workflow/worker.go
```diff
@@ -628,7 +628,11 @@ func (w *worker) runPurge(ctx context.Context, t *taskExecution) error {
 	if err != nil {
 		return err
 	}
-	return nil
+	// Just Publish an event that the deployment is successfully
+	// This event will stay into consul even if the deployment is actually purged...
+	// To prevent unexpected errors
+	_, err = events.PublishAndLogDeploymentStatusChange(ctx, t.cc.KV(), t.targetID, strings.ToLower(deployments.PURGED.String()))
```
laurentganne: So for clients providing unique names to the applications they ask to deploy (like fastML), there will be few entries leftover in consul never removed for each purge of such application.
Could we store such purged deployment IDs in consul with a timestamp of the purge operation, and the final entire cleanup would be done either if a user deploys an application with the same name like today, or if the purge has occured since a given time ?

loicalbertin: Ok so lets define an eviction timeout after which remaining events & logs of a purged deployment will be deleted. Same as before if a deployment with the same ID is redeployed before the eviction timeout then those logs & events will also be deleted.

PR #418, opened 2019-05-27
Bugfix/gh 417 non existent step

### LINKED ISSUE #417: Invalid memory address panic appends on a workflow with non-existent on success step reference
# Bug Report

## Description
Invalid memory address panic appends on a workflow with non-existent on success step reference

## Expected behavior
Explicit error describing the workflow issue
## Actual behavior
runtime error: invalid memory address or nil pointer dereference

## Priority

High

stebenoist:
# Pull Request description

## Description of the change
- Check if referenced wf step exist
- Return error when step doesn't exist 

* Invalid memory address panic appends on a workflow with non-existent on success step reference ([GH-417](https://github.com/ystia/yorc/issues/417))
## Applicable Issues
fixes #417

### DIFF
--- a/tasks/workflow/builder/builder.go
+++ b/tasks/workflow/builder/builder.go
@@ -99,7 +99,12 @@ func setCancelPath(s *Step) {
 }
 
 func buildStepFromWFStep(kv *api.KV, deploymentID, wfName, stepName string, wfSteps map[string]*tosca.Step, visitedMap map[string]*visitStep) (*Step, error) {
-	wfStep := wfSteps[stepName]
+	var wfStep *tosca.Step
+	var ok bool
+	if wfStep, ok = wfSteps[stepName]; !ok {
+		return nil, errors.Errorf("Referenced step with name:%q doesn't exist in the workflow:%q", stepName, wfName)
+	}
+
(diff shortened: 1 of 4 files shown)


tasks/workflow/builder/builder.go
```diff
@@ -99,7 +99,12 @@ func setCancelPath(s *Step) {
 }
 
 func buildStepFromWFStep(kv *api.KV, deploymentID, wfName, stepName string, wfSteps map[string]*tosca.Step, visitedMap map[string]*visitStep) (*Step, error) {
-	wfStep := wfSteps[stepName]
+	var wfStep *tosca.Step
+	var ok bool
+	if wfStep, ok = wfSteps[stepName]; !ok {
```
loicalbertin:
```suggestion
	if wfStep, ok := wfSteps[stepName]; !ok {
```

stebenoist: Hum...but wfStep is a new locale var to the if bloc (declared and not used) except if I am wrong ?

loicalbertin: Yep you are right

Issue #300, opened 2019-02-04
Yorc should support long standard operation names as well as short ones


stebenoist:
# Issue Type

Put an X between the brackets of the corresponding issue type

* [X] Bug report
* [ ] Documentation issue report
* [ ] Improvement request

## Description
Yorc only support Standard.create and not tosca.interfaces.node.lifecycle.Standard.create in operation handling. The short and the long version should be supported.
Idem for all tosca.interfaces.node.lifecycle operations (configure, start, stop, create, delete...)

PR #424, opened 2019-05-31
Support operation long naming

stebenoist:
# Pull Request description

### How I did it
 if operation name contains tosca.interfaces.node.lifecycle.standard:
-> use the short name "standard" to find related interface path in Consul KV
### How to verify it
- Pass U-Tests
- Create a Worflow with Alien4Cloud which contains a call-operation from standard interface
* Yorc should support long standard operation names as well as short ones ([GH-300](https://github.com/ystia/yorc/issues/300))

## Applicable Issues
#300

### DIFF
--- a/tosca/constants.go
+++ b/tosca/constants.go
@@ -25,3 +25,6 @@ const RunnableRunOperationName = RunnableInterfaceName + ".run"
 
 // RunnableCancelOperationName is the fully qualified name of the Cancel operation
 const RunnableCancelOperationName = RunnableInterfaceName + ".cancel"
+
+// StandardInterfaceName is the fully qualified name of the Standard interface
+const StandardInterfaceName = "tosca.interfaces.node.lifecycle.standard"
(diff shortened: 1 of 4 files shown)


tosca/constants.go
```diff
@@ -25,3 +25,9 @@ const RunnableRunOperationName = RunnableInterfaceName + ".run"
 
 // RunnableCancelOperationName is the fully qualified name of the Cancel operation
 const RunnableCancelOperationName = RunnableInterfaceName + ".cancel"
+
+// StandardInterfaceName is the fully qualified name of the Standard interface
+const StandardInterfaceName = "tosca.interfaces.node.lifecycle.standard"
+
+// StandardInterfaceShortName is the short name of the Standard interface
+const StandardInterfaceShortName = "standard"
```
loicalbertin: No need for a short version of runnable interface?
Is this not impacted by the short/long naming issue?

loicalbertin:
Should be backported into a 3.2 branch

tosca/constants.go
stebenoist: I don't find any usage of "short" version of runnable interface. So, right now, there is no need IMO of a short version of runnable interface.

loicalbertin: OK thanks for the clarification :+1:

loicalbertin: Well sorry but what about relationship operations `tosca.interfaces.relationships.configure`?
I think they are handled in the same way by Alien4Cloud.

stebenoist: Yes, I add this case

Issue #425, opened 2019-06-03
Expose bypass error parameter on workflow


loicalbertin:
**As:** An integrator
**I want to:** Have choice to bypass errors or not on workflows
**So that:** I can stop on errors for an undeploy workflow

**AC1:** Should be part of the REST API
**AC2:** Think if it should be limited to undeploy workflow
**AC3:** Default behavior should remain unchanged
**AC4:** Check that a4c plugin handle this case for undeployments
**AC5:** Sounds a good opportunity for increasing our test coverage on rest package

PR #434, opened 2019-06-06
Feature/gh 425 expose bypass error

stebenoist:
# Pull Request description

As: An integrator
I want to: Have choice to bypass errors or not on workflows
So that: I can stop on errors for an undeploy workflow

AC1: Should be part of the REST API
AC2: Think if it should be limited to undeploy workflow
AC3: Default behavior should remain unchanged
AC4: Check that a4c plugin handle this case for undeployments
AC5: Sounds a good opportunity for increasing our test coverage on rest package

### What I did
- Already exposed for custom workflows
- Add "stopOnError" query parameter handling to delete handler and store it via task data "continueOnError"
- Allow the worker to handle "continueOnError" on undeploy workflow
- Add option stop-on-error for CLI to use it

### How to verify it
- Deploy the welcome application
- Stop the python web server via custom command (this will make the undeploy firtst step fail)
- Undeploy it with CLI or REST client with stopOnError param
- Check the deployment status is UNDEPLOYMENT_FAILED and the undeployment is stopped

* Expose bypass error parameter on workflow ([GH-425](https://github.com/ystia/yorc/issues/425))
## Applicable Issues
#425

### DIFF
--- a/commands/deployments/dep_undeploy.go
+++ b/commands/deployments/dep_undeploy.go
@@ -81,7 +81,7 @@ func init() {
 
 	DeploymentsCmd.AddCommand(undeployCmd)
 	undeployCmd.PersistentFlags().BoolVarP(&purge, "purge", "p", false, "To use if you want to purge instead of undeploy")
-	undeployCmd.PersistentFlags().BoolVarP(&stopOnError, "stop-on-error", "", false, "By default if an error occurs during the undeployment, the error is bypassed and the un-deployment continues. This flag allows to stop if an error occurs.")
+	undeployCmd.PersistentFlags().BoolVarP(&stopOnError, "stop-on-error", "", false, "By default if an error occurs during the undeployment, the error is bypassed and the undeployment continues. This flag allows to stop if an error occurs.")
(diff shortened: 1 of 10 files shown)


commands/deployments/dep_undeploy.go
```diff
@@ -74,6 +81,7 @@ func init() {
 
 	DeploymentsCmd.AddCommand(undeployCmd)
 	undeployCmd.PersistentFlags().BoolVarP(&purge, "purge", "p", false, "To use if you want to purge instead of undeploy")
+	undeployCmd.PersistentFlags().BoolVarP(&stopOnError, "stop-on-error", "", false, "By default if an error occurs during the undeployment, the error is bypassed and the un-deployment continues. This flag allows to stop if an error occurs.")
```
laurentganne: Would need to document this new option `stop-on-error` in doc/cli.rst section `Undeploy a deployment`

stebenoist: Yes

stebenoist: Done

Issue #446, opened 2019-06-17
mem_per_node slurm option parameter is limited to integer number of GB


stebenoist:
# Bug Report

## Description
Should pass this param to string with slurm understandable units as [K|M|G|T]

## Priority

High

PR #447, opened 2019-06-17
Allow to provide units for mem_per_node job slurm option

stebenoist:
# Pull Request description

## Description of the change
Change the mem_per_node slurm option param from integer to string in order to provide slurm memory unit

### How to verify it
run a job with mem_per_node slurm job option to 300 M by instance
 mem_per_node slurm option parameter is limited to integer number of GB ([GH-446](https://github.com/ystia/yorc/issues/446))
## Applicable Issues
#446

### DIFF
--- a/data/tosca/yorc-slurm-types.yml
+++ b/data/tosca/yorc-slurm-types.yml
@@ -42,9 +42,11 @@ data_types:
         type: integer
         required: false
       mem_per_node:
-        type: string
-        description: The memory per node required to the job. Different units can be specified using the suffix [K|M|G|T].
+        type: scalar-unit.size
+        description: The memory per node required to the job.
         required: false
+        constraints:
+          - greater_or_equal: 0 KB
       time:
         type: string
         description: >
--- a/data/tosca/yorc-slurm-types.yml
+++ b/data/tosca/yorc-slurm-types.yml
@@ -42,8 +42,8 @@ data_types:
         type: integer
         required: false
       mem_per_node:
-        type: integer
-        description: The memory per node in GB required to the job.
+        type: string
+        description: The memory per node required to the job. Different units can be specified using the suffix [K|M|G|T].
         required: false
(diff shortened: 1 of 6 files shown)


data/tosca/yorc-slurm-types.yml
```diff
@@ -42,8 +42,8 @@ data_types:
         type: integer
         required: false
       mem_per_node:
-        type: integer
-        description: The memory per node in GB required to the job.
+        type: string
```
loicalbertin: isn't a [`scalar-unit.size`](http://docs.oasis-open.org/tosca/TOSCA-Simple-Profile-YAML/v1.2/TOSCA-Simple-Profile-YAML-v1.2.html#TYPE_TOSCA_SCALAR_UNIT_SIZE) a better type for this?

prov/slurm/execution.go
```diff
@@ -404,8 +402,8 @@ func (e *executionCommon) buildJobOpts() string {
 		opts += fmt.Sprintf(" --ntasks=%d", e.jobInfo.Tasks)
 	}
 	opts += fmt.Sprintf(" --nodes=%d", e.jobInfo.Nodes)
-	if e.jobInfo.Mem != 0 {
-		opts += fmt.Sprintf(" --mem=%dG", e.jobInfo.Mem)
+	if e.jobInfo.Mem != "" {
+		opts += fmt.Sprintf(" --mem=%s", e.jobInfo.Mem)
```
loicalbertin: I think using a `scalar-unit.size` will prevent using malformed values.
What do you think? 
That said, maybe a it will be necessary to canonize it using go-humanize for instance.

stebenoist: But I am ok it's more elegant and user-friendly in UI

loicalbertin: Cool :smile:

PR #448, opened 2019-06-18
Bugfix/gh 426 Wrong resources allocation on shareable Hosts Pool

### LINKED ISSUE #426: Wrong resources allocation on shareable Hosts Pool
# Bug Report

## Description

Attempting to deploy a topology containing a Compute node `ComputeA` and another Compute node `ComputeB`, with different components on them,
each Compute node having the constraint host.num_cpus set to 1 (needs 1 CPU),
on a Hosts Pool with shareable hosts, each having just 1 CPU.

Yorc is actually reserving the same Host for both Compute nodes `ComputeA` and `ComputeB`, updating the current available number of CPUs on this host to -1.

## Expected behavior

The expectation is that Yorc should have reserved two different hosts for `ComputeA` and `ComputeB`.

## Actual behavior

The same host is reserved for both `ComputeA` and `ComputeB` although it doesn't have enough resources to host applications to be installed on `ComputeA` and `ComputeB`.

## Output of `yorc version`

3.2.0

## Priority

Medium

laurentganne:
# Pull Request description

The orchestrator could allocate a shareable Host without enough resources to a deployment, in case of concurrent attempt to allocate resources.

### What I did

Added a unit test reproducing the issue.

Fixed the issue in  `prov/hostspool/executor.go` function `allocateHostToInstance()`:
previous code was allocating a host first, then later was updating this host resources labels.
Another worker could attempt to allocate a host in between, and could wrongly be allocated this shareable host event it didn't have enough resources left because the host resources labels were not yet updated.
Added a mutex to prevent other workers allocation attempts during the sequence host allocation / update of host resources label.

All other modifications are just done to fix sonar issues in already existing code: complexity, number of lines, similar code, too many parameters...

### How to verify it

Deploy on a Hosts Pool a topology containing a Compute node ComputeA and another Compute node ComputeB, with different components on them,
each Compute node having the constraint host.num_cpus set to 1 (needs 1 CPU),
on a Hosts Pool with shareable hosts, each having just 1 CPU.

Wrong resources allocation on shareable Hosts Pool ([GH-426](https://github.com/ystia/yorc/issues/426))

## Applicable Issues

Fixes #426


prov/hostspool/executor.go
```diff
@@ -33,67 +34,111 @@ import (
 	"github.com/ystia/yorc/v3/tosca"
 )
 
+const (
+	// endpointCapabilityName is the name of a TOSCA capability endpoint
+	endpointCapabilityName = "endpoint"
+	// publicAddressAttributeName is the name of an instance public address attribute
+	publicAddressAttributeName = "public_address"
```
loicalbertin: Just thinking but shouldn't we add at least this one as exported constants in the tosca package?

laurentganne: has yes, will do

prov/hostspool/executor.go
```diff
@@ -115,102 +160,153 @@ func (e *defaultExecutor) hostsPoolCreate(originalCtx context.Context, cc *api.C
...
-	instances, err := tasks.GetInstances(cc.KV(), taskID, deploymentID, nodeName)
+	instances, err := tasks.GetInstances(cc.KV(), op.taskID, op.deploymentID, op.nodeName)
 	if err != nil {
 		return err
 	}
+
+	return e.allocateHostToInstance(ctx, instances, shareable, filters, op, allocatedResources)
+}
+
+func (e *defaultExecutor) allocateHostToInstance(
```
adidanes: Maybe that this function's name could be _allocateHostsToInstances_ as for each instance the manager trys to allocate a host ?

laurentganne: ok will change

Issue #455, opened 2019-06-20
Job node state remains to "executing" when Ansible job fails


stebenoist:
# Bug Report

## Description
To reproduce it: 
- use attached archive
- deploy it
- run the wf run
- display info on node state
[job2.zip](https://github.com/ystia/yorc/files/3310526/job2.zip)

## Expected behavior
Job state is "error"
## Actual behavior
Job state is "executing"

## Priority

Medium
.

PR #456, opened 2019-06-20
Set job node state in error for Ansible job when failure occurs

stebenoist:
# Pull Request description

### How to verify it
See issue description
job node state remains to "executing" when Ansible job fails ([GH-455](https://github.com/ystia/yorc/issues/455))
## Applicable Issues
#455

### DIFF
--- a/prov/ansible/monitoring_job.go
+++ b/prov/ansible/monitoring_job.go
@@ -75,6 +75,7 @@ func (o *actionOperator) ExecAction(ctx context.Context, cfg config.Configuratio
 		case "RUNNING", "QUEUED":
 			return false, opErr
 		case "COMPLETED":
+			deployments.SetInstanceStateStringWithContextualLogs(ctx, consulutil.GetKV(), deploymentID, action.Data["nodeName"], "0", tosca.NodeStateStarted.String())
 			return true, opErr
 		case "FAILED":
 			if opErr == nil {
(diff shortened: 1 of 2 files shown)


prov/ansible/monitoring_job.go
```diff
@@ -78,6 +80,7 @@ func (o *actionOperator) ExecAction(ctx context.Context, cfg config.Configuratio
 			if opErr == nil {
 				opErr = errors.Errorf("job implementation of node %q was detected as failed", nodeName)
 			}
+			deployments.SetInstanceStateStringWithContextualLogs(ctx, consulutil.GetKV(), deploymentID, action.Data["nodeName"], "0", tosca.NodeStateError.String())
```
laurentganne: No need to call deployments.SetInstanceStateStringWithContextualLogs() when the state is completed ?

stebenoist: Yes, I think you're right

stebenoist: But which state ? OK or "completed" ?

laurentganne: It seems for Slurm Jobs, this is `Completed`, while for kubernetes jobs this is `Succeeded` and the failure is `Failed`

stebenoist: Ok it seems to be the specific states for job schedulers. But in our case, there is no job scheduler and we should just follow Tosca nodes states ? WYT ?

laurentganne: ok, started then ?

stebenoist: Yes

stebenoist: Done

PR #464, opened 2019-06-25
Convert scalar-unit mem size to KiB for Slurm option

stebenoist:
# Pull Request description

## Description of the change
Convert scalar-unit mem size to KiB for Slurm option
## Applicable Issues
#446

### DIFF
--- a/prov/slurm/helper.go
+++ b/prov/slurm/helper.go
@@ -407,13 +407,13 @@ func quoteArgs(t []string) string {
 	return args
 }
 
-// Slurm mem units are K|M|G|T ie KiB MiB GiB TiB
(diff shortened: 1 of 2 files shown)


stebenoist:
I reopened issue #446 for this fix. As we don't yet release anything, it can be part of the same changelog ? WYT ?

laurentganne:
> 
> 
> I reopened issue #446 for this fix. As we don't yet release anything, it can be part of the same changelog ? WYT ?

Ok, I was thinking to a user who would see this slurm option value change appear in logs, but we can consider  this is internal, and just have the interface change already documented in changelog

PR #470, opened 2019-06-27
Using now block storage API v3, still able to use v1 for tests purposes

### LINKED ISSUE #440: Support OpenStack Block storage API v3
Yorc is using the OpenStack Block Storage API v1 to create Block storage on demand.
According to https://developer.openstack.org/api-ref/block-storage/ the current version is v3, and even v2 is deprecated.

So, using a recent OpenStack installation, Yorc fails to create a block storage component with this error :
```
openstack_blockstorage_volume_v1.yorc-BlockStorage-0: 1 error(s) occurred:
openstack_blockstorage_volume_v1.yorc-BlockStorage-0: Error creating OpenStack volume: Resource not found
```

laurentganne:
# Pull Request description

### What I did

Using the OpenStack Block Storage API v3 instead of v1 not supported anymore (even v2 is deprecated).

To still be able to use Block Storage API v1 on old platforms for tests purposes, added a basic check:
checking if the User Domain mandatory required property (cf. https://developer.openstack.org/api-guide/quick-start/api-quick-start.html#authenticate ) is set in the Yorc OpenStack configuration.
If not set, considering this is an old platform on which Block Storage API v1 will be used.

### How I did it

Created a file  `prov/terraform/openstack/resource_types.go` where function `getOpenstackResourceTypes()` will return the API to use for the different OpenStack resources that can be created by Yorc.
Using the returned map `resource type` -> `api version` in other files in this package instead of using hard-coded strings.
Added a corresponding test file `prov/terraform/openstack/resource_types_test.go`.

All other changes are done to fix SonarCloud reported issues on existing code: complexity, number of lines in function, constants, test coverage.

### How to verify it

Bootstrap Yorc on an old OpenStack setup in insecure mode running (insecure mode to easily update the configuration later in the test):
```bash
$ ./yorc bootstrap -v inputs.yaml -n test --insecure
```

once done, login on the bootstrapped Alien4cloud, and edit the orchestrator configuration to add to the OpenStack location an on-demand BlockStorage resource of 10GB for example.
Create an application made of :
  * a compute nodes,
  * an attached block storage node,
  * a network attached to the compute node.

Deploy the application.
Login on the created compute instance.
Run this command:
```bash
$ lsblk
```
to check a disk `vdb` is attached to this compute instance.

Undeploy the application.

We will now redeploy this application on an new OpenStack Queens release.
For this, login on the Yorc host, edit `/etc/yorc/config.yorc.yaml` to change the openstack infrastructure configuration, to have something like :
```yaml
infrastructures:
  openstack:
    auth_url: http://1.2.3.4:5000/v3
    default_security_groups: [openbar, default]
    user_domain_name: Default
    user_name: xxx
    password: xxx
    project_name: myproject
    project_id: 36a1ae05995c4f099765c76825f3c30f
    private_network_name: private-xxx
    region: regionOne
    provisioning_over_fip_allowed: true
```

Then edit the yorc service `/etc/systemd/system/yorc.service` to replace one of the IP addresses listed in `NO_PROXY` env variable, by the IP address of your openstack setup, in this example `1.2.3.4`.
Then reload, stop then start the service:
```bash
systemctl daemon-reload
systemctl stop yorc
systemctl start yorc
```

Then login on Alien4Cloud, and change the orchestrator configuration on-demand compute instance resource to adapt the image id and the flavor id to values available on the new openstack setup :
```
image: 39ba04f8-2f2f-4721-943a-8c3752aac97b
flavor: a5fc3610-ff3f-4da9-a910-acbd634420de
```

Once done, deploy the application already created above, made of a Compute Node, an attached block storage node and a network.
Once done, login on the the created compute instance.
Run this command:
```bash
$ lsblk
```
to check a disk `vdb` is attached to this compute instance.

Support OpenStack Block storage API v3 ([GH-440](https://github.com/ystia/yorc/issues/440))

## Applicable Issues

Fixes #440

### DIFF
--- a/prov/terraform/openstack/generator.go
+++ b/prov/terraform/openstack/generator.go
@@ -37,6 +37,7 @@ import (
 
 const (
 	floatingIPEndpointCapAttribute = "/capabilities/endpoint/attributes/floating_ip_address"
+	consulKeysResource             = "consul_keys"
 )
 
 type osGenerator struct {
@@ -247,14 +248,14 @@ func (g *osGenerator) generateBlockStorageInfra(opts generateInfraOptions) error
 			Path:  path.Join(opts.instancesKey, opts.instanceName, "/attributes/volume_id"),
 			Value: fmt.Sprintf("${%s.%s.id}", opts.resourceTypes[blockStorageVolume], bsVolume.Name)}
 		consulKeys := commons.ConsulKeys{Keys: []commons.ConsulKey{consulKey}}
-		commons.AddResource(opts.infrastructure, "consul_keys", bsVolume.Name, &consulKeys)
+		commons.AddResource(opts.infrastructure, consulKeysResource, bsVolume.Name, &consulKeys)
 	} else {
 		name := opts.cfg.ResourcesPrefix + opts.nodeName + "-" + opts.instanceName
 		consulKey := commons.ConsulKey{
 			Path:  path.Join(opts.instancesKey, opts.instanceName, "/properties/volume_id"),
 			Value: strings.TrimSpace(bsIds[opts.instanceIndex])}
 		consulKeys := commons.ConsulKeys{Keys: []commons.ConsulKey{consulKey}}
-		commons.AddResource(opts.infrastructure, "consul_keys", name, &consulKeys)
+		commons.AddResource(opts.infrastructure, consulKeysResource, name, &consulKeys)
 	}
 
 	return err
@@ -311,7 +312,7 @@ func (g *osGenerator) generateFloatingIPInfra(opts generateInfraOptions) error {
 
 	}
 	consulKeys := commons.ConsulKeys{Keys: []commons.ConsulKey{consulKey}}
-	commons.AddResource(opts.infrastructure, "consul_keys", ip.Name, &consulKeys)
+	commons.AddResource(opts.infrastructure, consulKeysResource, ip.Name, &consulKeys)
 	return err
 }
 
@@ -350,7 +351,7 @@ func (g *osGenerator) generateNetworkInfra(opts generateInfraOptions) error {
 	consulKeys := commons.ConsulKeys{Keys: []commons.ConsulKey{consulKey}}
 	consulKeys.DependsOn = []string{fmt.Sprintf("%s.%s_subnet", opts.resourceTypes[networkingSubnet],
 		opts.nodeName)}
-	commons.AddResource(opts.infrastructure, "consul_keys", opts.nodeName, &consulKeys)
+	commons.AddResource(opts.infrastructure, consulKeysResource, opts.nodeName, &consulKeys)
 
 	return err
 }
--- a/prov/terraform/openstack/generator.go
+++ b/prov/terraform/openstack/generator.go
@@ -34,6 +34,10 @@ import (
 	"github.com/ystia/yorc/v4/tosca"
 )
 
+const (
+	floatingIPEndpointCapAttribute = "/capabilities/endpoint/attributes/floating_ip_address"
+)
+
 type osGenerator struct {
 }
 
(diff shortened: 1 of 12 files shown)


prov/terraform/openstack/generator.go
```diff
@@ -34,9 +35,27 @@ import (
 	"github.com/ystia/yorc/v4/tosca"
 )
 
+const (
+	floatingIPEndpointCapAttribute = "/capabilities/endpoint/attributes/floating_ip_address"
```
loicalbertin: Maybe not for this one as so many work is already in this PR but I'm not confortable with those consul storage specific stuff being out of the deployments package... We should consider refactor it somehow.

prov/terraform/openstack/generator.go
```diff
@@ -103,146 +108,250 @@ func (g *osGenerator) GenerateTerraformInfraForNode(ctx context.Context, cfg con
...
+	if err != nil {
+		return err
+	}
+
+	var consulKey commons.ConsulKey
+	if !ip.IsIP {
+		floatingIP := FloatingIP{Pool: ip.Pool}
+		commons.AddResource(opts.infrastructure, opts.resourceTypes[computeFloatingIP], ip.Name, &floatingIP)
+		consulKey = commons.ConsulKey{
+			Path:  path.Join(opts.instancesKey, opts.instanceName, floatingIPEndpointCapAttribute),
```
loicalbertin: Not sure but by doing this way, will we have attributes notifications works correctly?
Dealing with outputs and publishing attributes changes at the end of the process wouldn't be safer?
WDYT @stebenoist?

laurentganne:
> 
> I'm not comfortable with attributes changes notifications but I'm not sure we are doing them properly can you confirm?

Yes could verify the following node  instances attributes had the expected floating IP address value after a deployment:
  * Compute/0/attributes/public_address
  * Compute/0/attributes/public_ip_address
  * Compute/0/capabilities/endpoint/attributes/ip_address
  * FIPCompute/0/capabilities/endpoint/attributes/floating_ip_address

prov/terraform/openstack/osinstance.go
```diff
@@ -251,147 +377,192 @@ func (g *osGenerator) generateOSInstance(ctx context.Context, kv *api.KV, cfg co
...
+
+	prefix := path.Join(instancesKey, opts.instanceName, "attributes/networks", strconv.Itoa(i))
+	outputs[path.Join(prefix, "network_name")] = networkNameKey
+	outputs[path.Join(prefix, "network_id")] = networkIDKey
+	outputs[path.Join(prefix, "addresses")] = networkAddressesKey
+	return nil
+}
 
-	return errors.Errorf("Only one group requirement can be accepted for OpenStack compute with name:%q", nodeName)
+func getProperyValueString(kv *api.KV, deploymentID, nodeName, propertyName string) (string, error) {
```
stebenoist: typo on func name. Maybe, you can move this func into deployments package, nodes.go file ?

prov/terraform/openstack/generator.go
stebenoist: But I do'nt see anything wrong about it

prov/terraform/openstack/osinstance.go
laurentganne: ok, will change this

prov/terraform/openstack/generator.go
stebenoist: Nothing seems to have changed here

laurentganne: yes, just moved this code to a function, to fix a Sonar issue about the complexity of the main function

prov/terraform/openstack/osinstance.go
laurentganne: re_used deployments.GetStringNodeProperty()

Issue #473, opened 2019-07-02
Slurm Job cancel operation is failing during undeploy in some cases when purge task is done before


stebenoist:
# Bug Report

## Description
As the purge has been started, we got the following error:
TaskStep "yorc_automatic_cancellation_of_Job_run": error details: Missing mandatory parameter "type" for node "Job"
This bug is related to https://github.com/ystia/yorc/issues/460
## Expected behavior
Delay purge until all tasks are done/cancelled

## Actual behavior
Purge occurs despite running tasks

stebenoist:
If no cancel request is sent, add a timeout to cancel the task and stop purge delay

PR #475, opened 2019-07-04
Cancel running task after timeout when purge is delayed

stebenoist:
# Pull Request description

## Description of the change
Cancel running task after timeout when purge is delayed as no cancel request has been sent

## Applicable Issues
#473

### DIFF
--- a/tasks/workflow/worker.go
+++ b/tasks/workflow/worker.go
@@ -588,37 +588,34 @@ func (w *worker) runUndeploy(ctx context.Context, t *taskExecution) error {
 func (w *worker) delayPurge(ctx context.Context, t *taskExecution, taskID string) {
 	log.Debugf("let's delay purge as a task is still running:%q", taskID)
 	ticker := time.NewTicker(time.Second * 5)
+	defer ticker.Stop()
 	for {
 		select {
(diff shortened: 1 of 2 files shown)


tasks/workflow/worker.go
```diff
@@ -605,6 +605,20 @@ func (w *worker) delayPurge(ctx context.Context, t *taskExecution, taskID string
...
+			state, err := tasks.GetTaskStatus(t.cc.KV(), taskID)
+			if err != nil {
+				log.Printf("error occurred during delaying purge:%+v", err)
+				ticker.Stop()
+				return
+			}
+			if state == tasks.TaskStatusRUNNING {
+				log.Debugf("Purge delay timeout is reached: we cancel the task with ID:%q", taskID)
+				tasks.CancelTask(t.cc.KV(), taskID)
+				ticker.Stop()
```
loicalbertin: Maybe `ticker.Stop()` can be done in `defer` function this would easy maintenance :smile:

loicalbertin:
Changelog or not?

stebenoist:
I reopened the #473 issue so the changelog is already set

PR #184, opened 2018-10-23
Allow to run multiple workflows and custom commands in //

loicalbertin:
# Pull Request description

## Description of the change

Custom commands and custom workflows can be run along with other wf/cc but not if a deployment/undeployment/scaling workflow is currently running.

Now terraform and ansible recipes are stored under a path containing the execution task id.
By default those files are deleted after the execution by config options are available to keep them.

__Note:__ developed to be integrated on 3.0 and ported to 3.1 in another PR.

## Applicable Issues

Closes #182

PR #497, opened 2019-08-14
Should not be able to undeploy while a workflow is running

### LINKED ISSUE #460: Undeploying an application with a running workflow should not be possible
# Bug Report

## Description
Undeploying an application with a running workflow should not be possible.
The related workflow task should be canceled before undeploying the application.

## Expected behavior
Conflict error should be return by REST API when trying to undeploy an application with a running workflow.
## Actual behavior
It's possible to undeploy an application with a running workflow

## Priority

High / Medium / Low

laurentganne:
# Pull Request description

## Description of the change

Refusing to undeploy an application when a workflow is still running.

### What I did

A side effect of pull request https://github.com/ystia/yorc/pull/184 allowing to run multiple workflows and custom commands in parallel,
was that it allowed unexpectedly to undeploy an application while a custom command or custom workflow was running.
This pull request forbids again to perfom updates, undeployments, scaling of a deployment on which a custom command or custom workflow is running.

### How I did it

`tasks/tasks.go`
Added a slice argument `tasksTypesToIgnore` to function `TargetHasLivingTasks()`, to ignore living tasks whose type is in this slice of types to ignore.

`tasks/collector/collector.go`
In function `registerTask()`, added an argument `tasksTypesToIgnore` to call `TargetHasLivingTasks()`.
When registerTask is attempting to register a task of type `TaskTypeCustomCommand`  or `TaskTypeCustomWorkflow`, then we add both types to `tasksTypesToIgnore` to ignore such types.
This way it will possible to run multiple workflows and custom commands in parallel.
But when attempting to register an undeploy task for example, we don't add `TaskTypeCustomCommand` and `TaskTypeCustomWorkflow` to `tasksTypesToIgnore`, so if a custom workflow is running for the deployment, `registerTask()` will return an error.

`tasks/tasks_test.go`
Updated tests to take into account the new argument to `TargetHasLivingTasks()`. Added a test checking this argument allows to ignore types of tasks in progress.

### How to verify it

Using a yorc binary from this pull request,
import in Alien4Cloud catalog the tosca sample at https://github.com/laurentganne/tosca-samples/tree/master/iterations
Create an application from the imported topology template `ShortAndLongStartup`.
Edit the topology to remove components  `ShortStartupBash` and `LongStartupBash`, to just keep the component `LongStartupAnsible` on the compute instance.
Edit property values of this component `LongStartupAnsible`, to have `numberOFIteration` property set to 200, this will allow you to have an operation `iterate` that will perform 200 iterations, sleeping 1 second between each iteration.
So you will have an operation taking time to execute.

Deploy this application.
Once the application is deployed, in the deployed application `Workflow` view, select the workflow `iterate` and click on `Launch`.
Once it is launched, wait for a few seconds and click on `Undeploy` to attempt to undeploy the application.
There may be some UI refresh issue to display the error, so check Alien4Cloud logs at `/opt/alien4cloud/alien4cloud/logs/alien4cloud.log` and verify you have an error message about the undeployment:

```
INFO  [yorc-Yorc-task-4] FsmActions:250 - Undeploying yorc-essai-Environment
WARN  [yorc-Yorc-http-2] AsyncRestTemplate:527 - Async DELETE request for "http://10.0.0.212:8800/deployments/yorc-essai-Environment" resulted in 400 (Bad Request); invoking error handler
WARN  [yorc-Yorc-http-2] UndeployService:93 - Fail while Undeploying deployment [b533691c-c50b-41a2-a3ee-f8dd05504345] on orchestrator [7cd07327-fd76-44ac-826a-c7fd1d861689]
ERROR [yorc-Yorc-http-2] FsmActions:230 - HTTP Request KO : 400 Bad Request
```

Wait for the end of the workflow execution.
Click on `Undeploy` again, and check that the undeployment takes place now.

Undeploying an application with a running workflow should not be possible ([GH-460](https://github.com/ystia/yorc/issues/460))

## Applicable Issues

Fixes #460


laurentganne:
> 
> 
> Tested with Alien 2.2.0-SM6, alien4cloud-yorc-provider update branch and yorc 4.0.0.
> It works fine but have a synchronization pb in A4C GUI : whereas Undeploy is rejected with Bad Request, in the A4C Runtime view, the application's status gets stuck to "Undeployment in progress".

yes, workaround for the A4C UI refresh issue: select another tab, for example the tab `Adminstration`, then go back to your application, its state should be up to date now.

PR #513, opened 2019-09-11
Fix openstack network creation

### LINKED ISSUE #515: Invalid terraform reference in OpenStack network creation
# Bug Report

## Description
The creation of OpenStack networks is broken. In the subnet that gets created as well, the terraform reference used in the NetworkID is invalid which causes terraform to fail.

Define a topology with a private network and try to deploy it with yorc. The network creation will fail.

Only OpenStack related

## Priority
Medium

schrej:
## Description of the change
The creation of OpenStack networks was broken. The `generateSubnet` method needs the resource type for OpenStack networks in order to reference the network it should be attached to, but the resource type for subnets was passed.

### What I did
Fixed by passing the `networkingNetwork` type instead of `networkingSubnet`

Fixed a bug preventing OpenStack Networks from being created.

## Applicable Issues
fixes #515

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### BUG FIXES
 
-
+* Fixed a bug preventing OpenStack Networks from being created ([GH-515](https://github.com/ystia/yorc/issues/515))
 * A deployment may disappear from the deployments list while its currently running a purge task ([GH-504](https://github.com/ystia/yorc/issues/504))
 * A4C Logs are displaying stack error when workflow step fails ([GH-460](https://github.com/ystia/yorc/issues/503))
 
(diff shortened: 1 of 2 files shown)


loicalbertin:
Hi @schrej 

Thanks for contributing this fix.
Could you please open a bug report with the description you made into this PR?

We will review this soon.

schrej:
To make the review easier: The resourceType gets passed to [`getSubnetNetworkID`](https://github.com/ystia/yorc/blob/develop/prov/terraform/openstack/network.go#L149) where it's used to reference the network that will get created. In order for the network to be properly referenced, the resourceType needs to be Network, not Subnet, or it will reference itself.

PR #523, opened 2019-10-02
Fix possible overlap on generated batch wrappers

### LINKED ISSUE #522: Possible overlap on generated batch wrappers scripts when submitting several singularity jobs in parallel
# Bug Report

## Description

When not providing a batch script as part of the job (may be when using a command directly or for a singularity job) the actual srun command is wrapped into a generated batch script. The script is generated under the working directory (which is a property of a job) but with a fixed name `b.batch`. If submitting several jobs at the same time with the same working directory then the batch script may be overwritten by another job causing different issues (not running the good job, having an empty batch script,...).

## Expected behavior

Scripts should differs for different jobs

## Actual behavior

Name collisions causing different issues (not running the good job, having an empty batch script,...).

loicalbertin:
# Pull Request description

## Description of the change

Fix possible overlap on generated batch wrappers scripts when submitting several jobs in parallel

### What I did / How I did it

Generate a script name with a random UUID.
Now those scripts are deleted after submission, this was not the case up to now.  

* Fix possible overlap on generated batch wrappers scripts when submitting several singularity jobs in parallel ([GH-522](https://github.com/ystia/yorc/issues/522))

## Applicable Issues

Fixes #522

### DIFF
--- a/prov/slurm/execution.go
+++ b/prov/slurm/execution.go
@@ -27,6 +27,8 @@ import (
 	"strings"
 	"time"
 
+	"github.com/google/uuid"
+
 	"github.com/hashicorp/consul/api"
 	"github.com/pkg/errors"
 	"golang.org/x/sync/errgroup"
@@ -43,7 +45,7 @@ import (
 )
 
 const home = "~"
-const batchScript = "b.batch"
+const batchScript = "b-%s.batch"
 
 type execution interface {
 	resolveExecution() error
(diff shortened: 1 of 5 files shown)


prov/slurm/execution.go
```diff
@@ -426,25 +428,34 @@ func (e *executionCommon) prepareAndSubmitJob(ctx context.Context) error {
...
 	} else {
 		cmd = fmt.Sprintf("%s%ssbatch -D %s%s %s", e.addWorkingDirCmd(), e.buildEnvVars(), e.jobInfo.WorkingDir, e.buildJobOpts(), path.Join(e.jobInfo.WorkingDir, e.PrimaryFile))
 	}
 	return e.submitJob(ctx, cmd)
 }
 
-func (e *executionCommon) wrapCommand(innerCmd string) string {
-	pathScript := path.Join(e.jobInfo.WorkingDir, batchScript)
+func (e *executionCommon) wrapCommand(innerCmd string) (string, error) {
+	id, err := uuid.NewRandom()
```
laurentganne: Couldn't we use e.deploymentID  + e.nodeName + e.taskID + e.operation.Name ?
(not sure if there is this lack of entropy issue with uuid.newRandom() that we are seeing sometimes on Ansible)

prov/slurm/execution.go
```diff
@@ -426,25 +428,34 @@ func (e *executionCommon) prepareAndSubmitJob(ctx context.Context) error {
...
+	e.jobInfo.Artifacts = append(e.jobInfo.Artifacts, scriptName)
 	// Write script
 	cat := fmt.Sprintf(`cat <<'EOF' > %s
 #!/bin/bash
 
 %s
 EOF
 `, pathScript, innerCmd)
-	return fmt.Sprintf("%s%s%ssbatch -D %s%s %s", e.addWorkingDirCmd(), e.buildEnvVars(), cat, e.jobInfo.WorkingDir, e.buildJobOpts(), pathScript)
+	return fmt.Sprintf("%s%s%ssbatch -D %s%s %s; rm %s", e.addWorkingDirCmd(), e.buildEnvVars(), cat, e.jobInfo.WorkingDir, e.buildJobOpts(), pathScript, pathScript), nil
```
laurentganne: may use `/bin/rm %s` or `rm -f` just in case `rm` is aliases to `rm -i` on the host

loicalbertin: Yep good idea

prov/slurm/execution.go
loicalbertin: I think it will be more complicated to use  `e.deploymentID + e.nodeName + e.taskID + e.operation.Name`

1. This will be very very long as taskID is it self a UUID
2. We should deal with not file name friendly characters in node name and operation name

We already generate UUIDs so I think we have enough entropy.

PR #527, opened 2019-10-08
Fix BadAccess error in deployments.resolveValueAssignment

### LINKED ISSUE #526: deployments.resolveValueAssignment throws BadAccess if value can't be resolved
During writing a test that includes a relationship which includes a `get_attribute` reference I got a bad access error due to the attribute not being found.

## Expected behavior
No bad access error

## Actual behavior
Bad access error is thrown

I assume referencing an attribute that is not set and doesn't have a default value with the `get_attribute` function should trigger it.

## Priority
Low

I'm creating this Issue just for completeness, the fix is very obvious.

schrej:
This fixes a BadAccess error by passing the passed valueAssignment to the log call instead of a null pointer.
As the variable is checked for nil two lines above, this is very likely to be a typo.

fixes #526

### DIFF
--- a/deployments/resolver.go
+++ b/deployments/resolver.go
@@ -336,7 +336,7 @@ func resolveValueAssignment(kv *api.KV, deploymentID, nodeName, instanceName, re
 	}
 	if valueAssignment == nil {
 		ctx := events.NewContext(context.Background(), events.LogOptionalFields{events.NodeID: nodeName, events.InstanceID: instanceName})
-		events.WithContextOptionalFields(ctx).NewLogEntry(events.LogLevelWARN, deploymentID).Registerf("[WARNING] The value assignment %q hasn't be resolved for deployment: %q, node: %q, instance: %q. An empty string is returned instead", valueAssignment, deploymentID, nodeName, instanceName)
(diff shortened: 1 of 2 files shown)


deployments/resolver.go
```diff
@@ -336,7 +336,7 @@ func resolveValueAssignment(kv *api.KV, deploymentID, nodeName, instanceName, re
 	}
 	if valueAssignment == nil {
 		ctx := events.NewContext(context.Background(), events.LogOptionalFields{events.NodeID: nodeName, events.InstanceID: instanceName})
-		events.WithContextOptionalFields(ctx).NewLogEntry(events.LogLevelWARN, deploymentID).Registerf("[WARNING] The value assignment %q hasn't be resolved for deployment: %q, node: %q, instance: %q. An empty string is returned instead", valueAssignment, deploymentID, nodeName, instanceName)
+		events.WithContextOptionalFields(ctx).NewLogEntry(events.LogLevelWARN, deploymentID).Registerf("[WARNING] The value assignment %q hasn't be resolved for deployment: %q, node: %q, instance: %q. An empty string is returned instead", va, deploymentID, nodeName, instanceName)
```
schrej: the passed value assignment is parsed into `va` in line [328](https://github.com/ystia/yorc/pull/527/files#diff-434b86138c68a6e888c5e9568619124fR328)

PR #538, opened 2019-10-17
Added support of inline sbatch options & source env file

### LINKED ISSUE #537: Allow to provide inline configuration options for generated sbatch scripts
**As:** As a user
**I want to:** Add some inline options for generated sbatch scripts of SlurmJobs 
**So that:** I can provided some options that can't be provided otherwise like burst buffers configuration for instance

**AC1:** A list of options to be inserted into the genreated sbatch script could be provided

### LINKED ISSUE #541: Optionnaly source an environment file before submitting slurm jobs
**As:** A Slurm user
**I want to:** I want to have my custom environment loaded before submitting a Slurm job
**So that:** I can have specific customization within this file

**AC1:** Should be controlled in TOSCA
**AC2:** Should be optional

loicalbertin:
# Pull Request description

## Description of the change

Allow to provide extra options to Slurm for sbatch scripts specifically generated by Yorc to execute commands.

Also support sourcing of a user profile environment if any.

### What I did

Added a new TOSCA property for SlurmJob. Add use it for generating sbatch inline options.

* Allow to provide inline configuration options for generated sbatch scripts ([GH-537](https://github.com/ystia/yorc/issues/537))
* Optionnaly source an environment file before submitting slurm jobs ([GH-541](https://github.com/ystia/yorc/issues/541))

## Applicable Issues

Fixes #537 
Fixes #541

### DIFF
--- a/data/tosca/yorc-slurm-types.yml
+++ b/data/tosca/yorc-slurm-types.yml
@@ -90,6 +90,14 @@ data_types:
         required: false
         entry_schema:
           type: string
+      in_script_options:
+        type: list
+        description: |
+          List of options to be passed to sbatch as inline batch script options.
+          To be valid each element should start with a dash '#' character.
+        required: false
+        entry_schema:
+          type: string
 
 capability_types:
   yorc.capabilities.slurm.Endpoint:
--- a/data/tosca/yorc-slurm-types.yml
+++ b/data/tosca/yorc-slurm-types.yml
@@ -69,6 +69,8 @@ data_types:
         entry_schema:
           type: string
 
+# This type is backed by tosca.datatypes.SlurmExecutionOptions if modifying something here it should be reported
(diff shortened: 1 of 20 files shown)


data/tosca/yorc-slurm-types.yml
```diff
@@ -90,6 +92,14 @@ data_types:
         required: false
         entry_schema:
           type: string
+      in_script_options:
```
laurentganne: Shouldn't the `template_version` above be increased ? Wondering if we don't change it, if a yorc upgrade would store this new definition or would just keep the previous definition stored in consul

loicalbertin: Yes could be changed to 1.2.1 or 1.3.0 but anyway this do not correspond to what we now have in a4c as it is now always versioned along with a4c

Issue #479, opened 2019-07-09
Expose an API to manage locations


loicalbertin:
**As:** Yorc dev team
**I want to:** Manage locations using a REST API / CLI
**So that:** It can ease management/configuration of a Yorc cluster

**AC1:** Create/Update/Delete a location using a REST API
**AC2:** CLI commands should be available

PR #539, opened 2019-10-22
Feature/gh 479 locations api

adidanes:
# Pull Request description

## Description of the change

Enriched Yorc's http server with endpoints and handlers for **locations** management.

### What I did

Add endpoint implementations for CREATE/UPDATE/DELETE/LIST 

Add UpdateLocation method to the locations/Manager interface

### How to verify REST API
**List** locations :
$ curl --noproxy '*' -v -H "Content-Type: application/json" "http://localhost:8800/locations"

**Create** a new location:
$ curl --noproxy '*' -v -H "Accept: application/json" -H "Content-Type: application/json" -X PUT  "http://localhost:8800/locations/location4" --data '{"type":"test", "properties" : { "p1" : "v1", "p2" : "v2" }}'

Try also to create a location with a existent name

**Update** the created location:
$ curl --noproxy '*' -v -H "Accept: application/json" -H "Content-Type: application/json" -X PATCH  "http://localhost:8800/locations/location4" --data '{"type":"test1", "properties" : { "p1" : "v1", "p2" : "v2" }}'

Try also to update a non-existent location 

**Delete** a location (of any type):

$ curl --noproxy '*' -v -H "Accept: application/json" -H "Content-Type: application/json" -X DELETE "http://localhost:8800/locations/location4"

Try also to delete a non-existent location

### How to verify locations cli command
**List** locations :
$ ./yorc loc list

**Create** a new location:
$ ./yorc loc add --data '{"name: "location4", "type":"test", "properties" : { "p1" : "v1", "p2" : "v2" }}'

Try also to create a location with a existent name, or to create a location and forget to porovide its name.

**Update** the created location:
$ ./yorc loc update --data '{"name": "location4", "type":"test1", "properties" : { "p1" : "v1", "p2" : "v2" }}'

Try also to update a non-existent location 

**Delete** a location (of any type):

$ ./yorc loc delete location4"

Try also to delete a non-existent location

**Apply a new configuration file for locations:
$ ./yorc loc apply "/path/to/myLocConfig.json"

## Applicable Issues
https://github.com/ystia/yorc/issues/479


rest/http_api.md
```diff
@@ -1286,3 +1286,114 @@ Content-Type: application/json
...
+
+Other possible response response code is `400` if a location with the `<location_name>` does not exist.
+
+### Remove location
+
+Remove a location of a given type.
+
+'Content-Type' header should be set to 'application/json'.
+
+`DELETE /locations/<location_name>/<location_type>`
```
loicalbertin: I see no valid reasons why I should provide the location_type for deleting a location.

loicalbertin: Locations names should be unique so there is no risk of duplication

adidanes: The reason comes from the locations.Manager interface definition (RemoveLocation method).
I'll try to change this method signature and implementation if it's not convenient as it is now.

commands/locations/locations.go
```diff
@@ -0,0 +1,225 @@
...
+		return locationName, err
+	}
+	request.Header.Add("Content-Type", APPJSON)
+	response, err := client.Do(request)
+	if err != nil {
+		return locationName, err
+	}
+
+	//httputil.HandleHTTPStatusCode(response, "", "locations", http.StatusCreated)
+	httputil.HandleHTTPStatusCode(response, "", "locations", http.StatusOK)
```
laurentganne: The REST endpoint is returning http.StatusCreated, so attempting to add a location fails on error:
```
Error: Expecting HTTP Status code in [200] but got 201, reason "201 Created"
```

laurentganne: Fixed

commands/locations/locations_info.go
```diff
@@ -0,0 +1,56 @@
...
+			if err != nil {
+				httputil.ErrExit(err)
+			}
+			return getLocationInfo(client, args)
+		},
+	}
+	LocationsCmd.AddCommand(infoCmd)
+}
+
+func getLocationInfo(client httputil.HTTPClient, args []string) error {
```
laurentganne: Attempting to get a location info on a bootstrapped secured setup (credentials are stored in a vault, and location property values reference the secret in the vault) shows these warnings for example for an openstack location:
```
2019/11/22 15:27:27 [Warning] template parsing failed for key: "password" value: "{{ secret \"/secret/yorc/locations/myLocation\" \"data=password\" | print }}": template: GenericConfigMap:1: function "secret" not defined
2019/11/22 15:27:27 [Warning] template parsing failed for key: "tenant_name" value: "{{ secret \"/secret/yorc/locations/myLocation\" \"data=tenant_name\" | print }}": template: GenericConfigMap:1: function "secret" not defined
2019/11/22 15:27:27 [Warning] template parsing failed for key: "user_name" value: "{{ secret \"/secret/yorc/locations/myLocation\" \"data=user_name\" | print }}": template: GenericConfigMap:1: function "secret" not defined
```

rest/http_api.md
```diff
@@ -1288,3 +1288,114 @@ Content-Type: application/json
...
+```HTTP
+HTTP/1.1 200 OK
+Content-Type: application/json
+```
+
+```json
+{
+  "locations":[
+    {
+      "name":"location1",
```
laurentganne: Actually this call doesn't provide type and properties details for each location like shown in this example.
It provides a list of references, like:
```json
{"locations":[{"rel":"location","href":"/locations/myLocation1","type":"application/json"},{"rel":"location","href":"/locations/myLocation2","type":"application/json"}]}
```

And another endpoint should be documented to get a location description:
GET /locations/<locationName>

laurentganne:
Proposal to avoid template parsing warnings: [diffyorc.zip](https://github.com/ystia/yorc/files/3882697/diffyorc.zip)

commands/locations/locations_info.go
laurentganne: A solution to be discussed to avoid these warnings, could be to add a new function allowing to disable a template resolver, and the config tempalte resolver would be disabled here so it wouldn't try to resolve secret values.
Code diff of the proposal: [diffyorc.zip](https://github.com/ystia/yorc/files/3882697/diffyorc.zip)

laurentganne:
Can be done later, but it could be interesting to have a CLI option to export locations details to a yaml or json file, so that the user can export the current locations definition in a file, modify the file and then apply it, to make locations changes, similarly to what can be done with hosts pools.
Something like:

**yorc locations export**
Flags:
* **--output** or **-o**: Output format, **yaml** or **json** (default **yaml**)
 * **--file** or **-f**: Path to a file where to store the output (default standard output)

doc/cli.rst
```diff
@@ -294,6 +294,29 @@ Flags:
...
+
+For brevity ``locations`` supports the following aliases: ``locs``, ``loc`` and ``l``.
+
+List locations
+~~~~~~~~~~~~~~
+
+This command allows to list all the locations with their complete definition: type and properties.
+
+.. code-block:: bash
+
```
laurentganne: Missing description for commands apply, add, delete, update, info

Issue #530, opened 2019-10-09
Refactor deployments package to be able to use different storage backends


loicalbertin:
**As:** A Yorc developer 
**I want to:** Refactor the `deployments` API to make it less dependent to Consul allowing to change the backend
**So that:** We will be able to use different backends which can be an answer to #360

**AC1:** API surface should not anymore depend on consul
**AC2:** API business logic should be decoupled from the storage backend
**AC3:** Think about our best practice (always pass context, ...)
**AC4:** Think about generalizing usage of struct representing your data type just like here: https://github.com/ystia/yorc/blob/develop/prov/slurm/helper.go#L107
with datatypes defined here: https://github.com/ystia/yorc/blob/develop/tosca/datatypes/credential.go

PR #540, opened 2019-10-22
Bastion hosts support for prov/ansible and prov/terraform

### LINKED ISSUE #128: Support provisioning through an ssh "bastion"
As: an operator
I want: be able to provision nodes even if they are not accessible directly through ssh
So that: I can be able to deploy an application in an environment that do not support direct ssh access

AC1: Use ssh bastion for Ansible
AC2: Use this kind of bastion for terraform when checking if the node is available

schrej:
This PR adds support for provisioning unreachable compute instances within private networks by leveraging bastion hosts which are reachable by yorc.

I'm aware this is a large PR and I'll try to explain everything below. If there are any further questions, please ask here or reach out on Gitter.

### How I did it
First of all, there are changes to yorc's tosca types: There is a new SoftwareComponent as well as a relationship. Additionally, `yorc.capabilities.Endpoint.ProvisioningAdmin` was extended to support configuring a bastion host to use.

The `prov/terraform` package now uses the bastion host in order to perform connection checks after creating an instance. This is done using the built in [configuration of bastion hosts](https://www.terraform.io/docs/provisioners/connection.html#connecting-through-a-bastion-host-with-ssh) that terraform provides.
The functionality is implemented in `prov/terraform/commons` and should work for all supported cloud platforms, only OpenStack has been tested so far, though.

The `prov/ansible` package also supports bastion hosts. As Ansible doesn't offer native support for bastion hosts, a config option is used to add parameters to the ssh command that is used by ansible. They add a ssh command as `ProxyCommand` and disable host key checking, as the interactive prompt to accept a new host key will cause ansible to fail.
We may have to think about a better solution here to improve security.

Both packages leverage helper functions added to `helper/provutil`. They allow to retrieve the bastion host configuration for a given instance. The configuration in the ProvisioningAdmin takes precedence over the bastion host relationship.

### How to verify it
Tests are implemented and should cover all changes. There are no specific tests for the different platforms in `prov/terraform`. Instead, the `commons` package is tested.

A proper e2e test will probably require `provisioning_over_fip_allowed: true` to be enabled for the OpenStack configuration. Alternatively, it can also be tested with private IPs, of course.
Unfortunately I haven't tested that part with the current implementation of OpenStack networking. We are using my re-implementation that's available in #518. I will work on a proper PR for that, but it might introduce breaking changes to the behavior of the networking in the OpenStack package. It should be closer to the tosca standard though.

As we are only testing yorc in combination with a4c, here is a definitions archive with the software component and the relationship: [bastion-test-types.zip](https://github.com/ystia/yorc/files/3756931/bastion-test-types.zip)
Those probably need to be added to the types provided by the alien4cloud-yorc-provider.

For testing the relationship, the current types provided by the a4c-y-p need to be modified. As I haven't figured out how to do that properly, and just got it to work by chance at some point, the changes can be found [here](https://github.com/OpenTelekomCloud/yorc/commit/7294324b549601720c41ac2c186bc5750a033f58#diff-0b7cc1b08f1c627c98286ed53b8aa288R39-R94).
I've used the topology in the following diagram for testing, which you can find [here](https://gist.github.com/schrej/9c2fe8c8669927151331a4cfb4f4ff67).
image
I haven't e2e tested the configuration in the `AdminEndpoint` yet, but will do so soon. It *should* work though, it's using the same implementation within the providers as the relationship.

### Limitations
It's not possible to chain bastion hosts. While it would potentially possible by forwarding ports with ssh manually, I don't think it's a sensible use case.

The `prov/ansible` implementation doesn't support using passwords for the bastion host as you can't (and probably shouldn't) pass passwords as an argument to ansible, nor can you insert it in the interactive prompt of ssh. Ansible will just fail when the interactive prompt comes up.

Even though the relationship includes `network_name` and `port_name`, they are currently not used. Instead, only the `ip_address` determines which ip to use to reach the bastion host. This is mostly due to the (imho) somewhat incomplete implementation of networking in the openstack package (and maybe the others as well). This should probably be improved.

### Notes
I've renamed the `tosca/datatypes` package to `tosca/types` during development, as I added additonal types that aren't `tosca.datatype.*`s. I removed them again later but decided to keep the name as it makes more sense imho, also in preparation for maybe using more static types throughout yorc. I'm open to reverting it if you don't want that kind of random change.

### Thoughts/Room for improvements
Currently, a fixed software component (`yorc.nodes.SSHBastionHost`) is used and required by the relationship. As the name indicates it's a yorc provided type, this should be provided, which would require someone to keep it up to date manually, ideally support multiple OSes and harden the OS automatically and so on.
Also, the customers might want to use their own implementation of a bastion host (because of e.g. security requirements etc.).
I think a SSHBastionHost capability would be a better solution, and wen can then provide an example in the forge repository, but I haven't manged to get it to work. It might be because I lack experience with tosca so any help/ideas is much appreciated!

## Applicable Issues
resolves #128


adidanes:
[bastion-test-Environment-0.1.0-SNAPSHOT.log](https://github.com/ystia/yorc/files/3784134/bastion-test-Environment-0.1.0-SNAPSHOT.log)

prov/ansible/execution.go
```diff
@@ -452,6 +453,7 @@ func (e *executionCommon) resolveHostsOnCompute(nodeName string, instances []str
 					ipAddressStr := config.DefaultConfigTemplateResolver.ResolveValueWithTemplates("host.ip_address", ipAddress.RawString()).(string)
 					instanceName := operations.GetInstanceName(nodeName, instance)
 					hostConn := &hostConnection{host: ipAddressStr, instanceID: instance}
+					hostConn.bastion, err = provutil.GetInstanceBastionHost(e.ctx, e.kv, e.deploymentID, host)
```
stebenoist: an err check is missing on provutil.GetInstanceBastionHost call

schrej:
Please don't merge this yet. I want to switch from a SoftwareComponent to a Capability.

stebenoist:
Ok, we let you merge when it's OK.

If we valid #530 before you mege, be careful to the changes this PR contains. Conflicts should be easy to resolve but I can help you !

schrej:
I've now switched from using a SoftwareComponent to a Capability. The relationship now requires the `yorc.capabilities.SSHBastionHost` as an endpoint. That way yorc no longer needs to provide a SoftwareComponent and users have to define their own. This way it's also easier to update the component. If you want to test, here is an example component: [bastion-test-types.zip](https://github.com/ystia/yorc/files/3824817/bastion-test-types.zip)
I've updated the topology accordingly, you can still find it [here](https://gist.github.com/schrej/9c2fe8c8669927151331a4cfb4f4ff67).

adidanes:
Hi @schrej,
I re-tested the PR and validated the usage of `yorc.capabilities.SSHBastionHost` capability.
It works fine. So from my point of view, PR is ready for merge.

schrej:
We've also been using it since about two weeks and haven't found any further issues so far, at least with ansible and openstack.

helper/provutil/bastion.go
```diff
@@ -90,57 +91,61 @@ func getInstanceBastionHostFromEndpoint(ctx context.Context, deploymentID, nodeN
 	return b, nil
 }
 
-func getInstanceBastionHostFromRelationship(ctx context.Context, deploymentID, nodeName, depIdx string) (*sshutil.BastionHostConfig, error) {
-	bastionNodeName, err := deployments.GetTargetNodeForRequirement(ctx, deploymentID, nodeName, depIdx)
+func getInstanceBastionHostFromRelationship(ctx context.Context, deploymentID, nodeName, reqIdx string) (b *sshutil.BastionHostConfig, err error) {
+	bastNode, err := deployments.GetTargetNodeForRequirement(ctx, deploymentID, nodeName, reqIdx)
 	if err != nil {
-		return nil, err
+		return
```
stebenoist: seems weird not to return error here and below ?

loicalbertin: https://github.com/golang/go/wiki/CodeReviewComments#named-result-parameters

schrej: Hmm, makes sense, I've removed them.

schrej:
Is there anything still preventing this getting merged?

Makefile
```diff
@@ -58,7 +58,7 @@ build: test
 	 "
 	 @rm -f ./build/embeddedResources.zip
 	 @cd ./commands && zip -q -r ../build/embeddedResources.zip ./bootstrap/resources/*
-	 @cd ./data && zip -q -r ../build/embeddedResources.zip ./*
+	 @cd ./data && zip -q -r ../build/embeddedResources.zip ./* -x tosca/test-types.yml
```
schrej: TIL that the `data/tosca` definitions get added to the binary and stored to consul during startup. This ignores the test-types during zip compression so they don't get added.

stebenoist: Thanks ! Good to know !

loicalbertin:
Hi @schrej 

could you please elaborate a bit more on:

> I've renamed the tosca/datatypes package to tosca/types during development, [...] as it makes more sense imho, also in preparation for maybe using more static types throughout yorc. I'm open to reverting it if you don't want that kind of random change.

What was wrong with datatypes name in your opinion?
For me, `TOSCA datatypes` are special kind of `TOSCA types` just like nodes types or capabilities types. The idea of this package was to contains Go structs representations allowing to serialize/deserialize TOSCA value assignments (ie: properties/attributes values).

As a side note we implemented a tool https://github.com/ystia/tdt2go that generates Go structs from TOSCA datatypes. This will be shipped on https://github.com/ystia/yorc/compare/technical/GH-530-storage-refactoring (PR not create for now).

Makefile
loicalbertin: I do not get the point on having this for testing only :thinking: 
If it is for testing it is probably not the right place

schrej:
> could you please elaborate a bit more on:
> What was wrong with datatypes name in your opinion?
Nothing is wrong with it. At some point I was adding other tosca types to it that weren't datatypes. Instead of creating additional packages I just renamed it to types. My implementation made no real sense, so I removed the types again, but I thought a single package made more sense so I kept the name change.
If you'd prefer to have packages for the different kinds of tosca types, e.g. `nodes`, `datatypes`, `capabilites` etc., I'll revert that change.

> As a side note we implemented a tool https://github.com/ystia/tdt2go that generates Go structs from TOSCA datatypes. This will be shipped on https://github.com/ystia/yorc/compare/technical/GH-530-storage-refactoring (PR not create for now).
Very interesting. I'd really like to have more types within yorc. I started working on a function to retrieve multiple/all properties of a node by just passing a reference to an array, similar to the json decoder. It currently parses the json struct tag using reflection and tries to fill all fields in the struct. It makes it a lot easier and way less verbose to retrieve multiple properties and they are also available in a proper structure.

loicalbertin:
> > could you please elaborate a bit more on:
> > What was wrong with datatypes name in your opinion?
> 
> Nothing is wrong with it. At some point I was adding other tosca types to it that weren't datatypes. Instead of creating additional packages I just renamed it to types. My implementation made no real sense, so I removed the types again, but I thought a single package made more sense so I kept the name change.
> If you'd prefer to have packages for the different kinds of tosca types, e.g. `nodes`, `datatypes`, `capabilites` etc., I'll revert that change.

Sorry but except for datatypes which are a kind of meta-definition within TOSCA I can't figure out what will need to be serialized/de-serialized with custom structs that are not already in the `tosca` package?

> > As a side note we implemented a tool https://github.com/ystia/tdt2go that generates Go structs from TOSCA datatypes. This will be shipped on https://github.com/ystia/yorc/compare/technical/GH-530-storage-refactoring (PR not create for now).
> 
> Very interesting. I'd really like to have more types within yorc. I started working on a function to retrieve multiple/all properties of a node by just passing a reference to an array, similar to the json decoder. It currently parses the json struct tag using reflection and tries to fill all fields in the struct. It makes it a lot easier and way less verbose to retrieve multiple properties and they are also available in a proper structure.
> If you are interested in improvements in that general direction I'm happy to contribute. But maybe we should continue discussions about that on Gitter, as it's rather offtopic here.

Ok

schrej:
> Sorry but except for datatypes which are a kind of meta-definition within TOSCA I can't figure out what will need to be serialized/de-serialized with custom structs that are not already in the `tosca` package?

I put the bastion host capability in there at one point for example. At the time it wasn't of any use though, as there was no easy way to fill a struct with the properties of a node. Basically anything that is not provider specific could go in there (datatypes, policies, capabilities, ...) in my opinion once multiple properties can be queried at once using a tagged struct.

loicalbertin:
Ok sounds good to me.

Just one more thing. I would prefer having `data/tosca/test-types.yml` into `helper/provutil/testdata/test-types.yml` and keep `data/tosca` for things that are builtin Yorc.

loicalbertin:
@schrej PR merged thanks for this great contribution!

PR #548, opened 2019-11-18
Version exposed in the API

hilde2:
# Pull Request description

### How to verify it
Just GET http://localhost:8800/server/info with a rest client like insomnia.

### DIFF
--- a/commands/version.go
+++ b/commands/version.go
@@ -18,16 +18,15 @@ import (
 	"fmt"
 
 	"github.com/spf13/cobra"
-
+	"github.com/ystia/yorc/v4/config"
 	"github.com/ystia/yorc/v4/helper/consulutil"
 )
-
-var version = "unknown phantom version"
 
 var gitCommit = "dev version"
 
 func init() {
 	var quiet bool
+	var version = config.YorcVersion
 	versionCmd := &cobra.Command{
 
 		Use:   "version",
--- a/commands/version.go
+++ b/commands/version.go
@@ -18,15 +18,15 @@ import (
 	"fmt"
 
 	"github.com/spf13/cobra"
-	"github.com/ystia/yorc/v4/config"
 	"github.com/ystia/yorc/v4/helper/consulutil"
+	"github.com/ystia/yorc/v4/server/info"
 )
-
-var gitCommit = "dev version"
 
 func init() {
 	var quiet bool
-	var version = config.YorcVersion
+	var version = info.YorcVersion
+	var gitCommit = info.GitCommit
+
 	versionCmd := &cobra.Command{
 
 		Use:   "version",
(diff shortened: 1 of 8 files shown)


config/info.go
```diff
@@ -0,0 +1,18 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package config
+
+// YorcVersion is the used version of YORC
+var YorcVersion = "Must Be defined by MakeFile"
```
loicalbertin: Hum not sure that the `config` package is the right place for this. `config` is for things that are configurable...

commands/version.go
```diff
@@ -18,16 +18,15 @@ import (
...
 
 	"github.com/spf13/cobra"
-
+	"github.com/ystia/yorc/v4/config"
 	"github.com/ystia/yorc/v4/helper/consulutil"
 )
 
-var version = "unknown phantom version"
-
 var gitCommit = "dev version"
```
loicalbertin: I would keep the `version` and `gitCommit` in the same place they describe the same thing.

rest/http.go
```diff
@@ -143,6 +143,7 @@ func NewServer(configuration config.Configuration, client *api.Client, shutdownC
 
 func (s *Server) registerHandlers() {
 	commonHandlers := alice.New(telemetryHandler, loggingHandler, recoverHandler)
+	s.router.Get("/info/yorcVersion", commonHandlers.Append(acceptHandler("application/json")).ThenFunc(s.getInfoHandler))
```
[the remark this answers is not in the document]
loicalbertin: I'm not sure that `/info` is the best endpoint path.
Maybe we should have `/server/info` this way we will be able to add some other endpoints to control the server under the  ̀/server` path.
@ystia/yorc-dev WDYT?

Also we may move `/health` under `/server/health` as it is giving info on the local server only. @ystia/yorc-dev WDYT?

loicalbertin:
Also please have a look on Sonar issues :smile:

commands/version.go
loicalbertin: `gitCommit`should somehow be part of the version information  retrieved by the API

config/info.go
hilde2: I didn't find any better fitting package for this. `config` looked liked the the least worst. 
So maybe a new one must be created ? Maybe `info` or `others`

loicalbertin: And what about a `server/info` package for info about a server? 
@ystia/yorc-dev WDYT?

rest/http.go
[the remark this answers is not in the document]
stebenoist: Good idea !

config/info.go
stebenoist: Yep

PR #576, opened 2019-12-12
Feature/gh 122 aws volume attachment

### LINKED ISSUE #122: Volume attachment on AWS Compute nodes
As: An application designer
I want to: Be able to attach a volume to a Compute
So that: I can have more reliable deployments

AC1: Same level of functionalities than on openstack
AC2: If a volume does not exist it should be created
AC3: We should be able to attach to an pre-existing volume

HildericSB:
# Pull Request description
## Description of the change
Support of EBS(Elastic Block Store) and attachment of EBS to an AWS instance.

### What I did
- Refactoring of "GenerateTerraformInfraForNode()" of generator.go so it can handle the generations of multiples intances correctly. 
- Create "generateEBS()" that can create an EBS with the correct proprerties (zone, size, encryption) and with terraform outputs (volumeid, arn)
- "addAttachedDisks()"  added in aws_instance. It handle the volume attachment by creating a "aws_volume_attachment" ressource.
- Created a nodeParams struct to reduce the number of args in some functions.

### How to verify it
Create a topolgy that have a compute and a block storage.

!! The device name of the block storage must be the same as the name on the instance. !!
/dev/sdf could become /dev/xvdf on most recent Linux system.
[see](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/device_naming.html) 

## Applicable Issues
#Fixes #122


prov/terraform/commons/utils.go
```diff
@@ -12,16 +12,16 @@
...
 import (
 	"crypto/sha1"
 	"fmt"
-
 	"github.com/ystia/yorc/v4/config"
 )
 
-func getResourcesPrefix(cfg config.Configuration, deploymentID string) string {
+// GetResourcesPrefix : TODO
+func GetResourcesPrefix(cfg config.Configuration, deploymentID string) string {
```
loicalbertin: For me this function is specific to google.
We added this because Google Cloud does not allow to have resources with the same name.
This limitation doesn't exist on OpenStack & AWS.

HildericSB: Oh ok ! I wasn't sure, so I've done it just in case. I'll rollback.

PR #579, opened 2019-12-18
Backport premium test issue and minor bug from premium sources

stebenoist:
- Backport from premium
- Update changelog with breaking changes in deployments API (missed from https://github.com/ystia/yorc/pull/568)
- Fix minor issues


plugin/config.go
```diff
@@ -54,7 +55,9 @@ func (cm *defaultConfigManager) SetupConfig(cfg config.Configuration) error {
 		maxPubSub = config.DefaultConsulPubMaxRoutines
 	}
 	consulutil.InitConsulPublisher(maxPubSub, kv)
-	return nil
+
+	// Load stores in order to allow plugins to access different stores
+	return storage.LoadStores(cfg)
```
loicalbertin: How this will work?

Should delegate/operation plugins themselves be compiled in plugin mode to have access to stores plugins?

stebenoist: Yes, you need to compile delegate/operations plugins with CGO_ENABLED=1

Issue #584, opened 2020-01-22
Unable to get network ID attribute for Openstack many compute instances


stebenoist:
# Bug Report

## Description

To retrieve netwokl_id attribute for Openstack compute instance, the related instance name used corresponds to the compute instance instead of network instance which must be "0"
https://github.com/ystia/yorc/blob/develop/prov/terraform/openstack/osinstance.go#L467

PR #587, opened 2020-01-22
Use "0" for default instance name for Openstack network

stebenoist:
# Pull Request description

## Description of the change
Replace compute instance name by network instance name (as "0" as network is not scalable) for retrieveing network_id attribute for network node

* Unable to get network ID attribute for Openstack many compute instances ([GH-584](https://github.com/ystia/yorc/issues/584))
## Applicable Issues
#584

### DIFF
--- a/prov/terraform/openstack/osinstance.go
+++ b/prov/terraform/openstack/osinstance.go
@@ -460,33 +460,14 @@ func computeNetworkAttributes(ctx context.Context, opts osInstanceOptions,
 	instance *ComputeInstance, outputs map[string]string) error {
 
 	log.Debugf("Looking for Network id for %q", networkNodeName)
-	var networkID string
-	resultChan := make(chan string, 1)
-	go func() {
-		for {
-			nID, err := deployments.GetInstanceAttributeValue(ctx, opts.deploymentID, networkNodeName, opts.instanceName, "network_id")
-			if err != nil {
-				log.Printf("[Warning] bypassing error while waiting for a network id: %v", err)
-			}
-			// As networkID is an optional property GetInstanceAttribute then GetProperty
-			// may return an empty networkID so keep checking as long as we have it
-			if nID != nil && nID.RawString() != "" {
-				resultChan <- nID.RawString()
-				return
-			}
-			select {
-			case <-time.After(1 * time.Second):
-			case <-ctx.Done():
-				// context cancelled, give up!
-				return
-			}
-		}
-	}()
-	select {
-	case networkID = <-resultChan:
-	case <-ctx.Done():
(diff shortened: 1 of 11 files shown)


prov/terraform/openstack/osinstance.go
```diff
@@ -460,33 +460,14 @@ func computeNetworkAttributes(ctx context.Context, opts osInstanceOptions,
...
-			}
-		}
-	}()
-	select {
-	case networkID = <-resultChan:
-	case <-ctx.Done():
-		return ctx.Err()
+
+	// As Network instance is unique
+	defaultNetworkInstanceName := "0"
```
schrej: I would implement this as a global variable (e.g. `DefaultInstanceName` or `FirstInstanceName`) in the deployments package instead, in case [this implementation](https://github.com/ystia/yorc/blob/develop/deployments/definition_store.go#L585) ever changes.

prov/terraform/openstack/osinstance.go
```diff
@@ -460,33 +460,14 @@ func computeNetworkAttributes(ctx context.Context, opts osInstanceOptions,
...
-			case <-ctx.Done():
-				// context cancelled, give up!
-				return
-			}
-		}
-	}()
-	select {
-	case networkID = <-resultChan:
-	case <-ctx.Done():
-		return ctx.Err()
```
schrej: Can't we also remove this polling logic for floating IPs as well?

prov/terraform/openstack/osinstance.go
stebenoist: Sounds good to me

prov/terraform/openstack/osinstance.go
stebenoist: Yes

prov/terraform/openstack/osinstance.go
```diff
@@ -414,28 +383,12 @@ func computeFloatingIPAddress(ctx context.Context, opts osInstanceOptions,
...
-				// context cancelled, give up!
-				return
-			}
-		}
-	}()
-	select {
-	case floatingIP = <-resultChan:
-	case <-ctx.Done():
-		return ctx.Err()
+	floatingIP, err := deployments.LookupInstanceCapabilityAttributeValue(ctx, deploymentID, networkNodeName, deployments.DefaultInstanceName, "endpoint", "floating_ip_address")
```
schrej: For floating ip addresses we will need to use `instanceName`, as there is one FIP per compute instance.

stebenoist: Yes, I think you're right/ I fix it too

PR #621, opened 2020-03-24
Check for location type mismatch

### LINKED ISSUE #550: Check for location type mismatch
# Bug Report
Yorc should check that the type of the location requested for the deployment of an application matches the type defined in the yorc configuration.
In case of mismatch, deploymeny should fail and the error log should inform the user about the mismatch.
 
## Description
An application template specifies for each node the requested location name. The location type is not specified, but the node type contains the requested infrastructure type (openstack, aws, etc).
If in the Yorc configuration a location with the given name exists, but its type does not correspond to the requested infrastructure, the deployment will fail as configuration properties are possibly incoherent.

## Expected behavior
Yorc should check that the requested location type is coherent with the requested infrastructure.

## Actual behavior
Yorc tries to deploy application to the requested infrastructure but the location configuration does not correspond to the requested one.

Try to deploy a simple application (one compute node) with A4C & Yorc.
The A4C Yorc orchestrator configuration contains a location **L** of type T1. 
The Yorc server is configured with a location **L** of type T2. 
1. Create a simple application (one compute node) in A4C.
2. Choose location **L** to deploy it.
3. Check logs, you can see that Yorc tries to deploy using an infrastructure provider of type T1 configured with properties corresponding to a T2 type infrastructure.

In my test T1 is AWS and T2 is OpenStack. 
The error message is : provider.was: Not a valid region: RegionOne 
This happens because in the Yorc configuration, L location has has 'type' set to 'openstack' and a property 'region' set to 'RegionOne'.

## Output of `yorc version`
4.0.0

## Priority
Low

adidanes:
# Pull Request description

## Description of the change
Detect that a location targeted by a deployment has the expected type.

### What I did
If the targeted location exists, check that its type is the same as the requested infrastructure type.

### How to verify it
Request deployments on a location `locName` corresponding to a given infrastructure type `infraType`. 
Define in yorc a location  `locName` having type `locType` different from `infraType`

Check that deployment fails with the following error message : 
"The location  `locName` has  type `locType` and not the expected `infraType` type"

Check for location type mismatch ([GH-550](https://github.com/ystia/yorc/issues/550))

## Applicable Issues

Closes #550

### DIFF
--- a/locations/location_mgr.go
+++ b/locations/location_mgr.go
@@ -194,9 +194,8 @@ func (mgr *locationManager) GetLocationProperties(locationName, locationType str
 	// got configuration for the requested location
 	props = lConfig.Properties
 	// 	check if no mismatch between the type defined in the configuration and the given location type
-	kvLocationType := lConfig.Type
-	if locationType != kvLocationType {
-		return props, errors.Errorf("The location %q has %s type defined and not the requested %s type", locationName, kvLocationType, locationType)
+	if locationType != lConfig.Type {
+		return props, errors.Errorf("The location %q has %s type defined and not the requested %s type", locationName, lConfig.Type, locationType)
 	}
 
 	return props, err
--- a/locations/location_mgr.go
+++ b/locations/location_mgr.go
@@ -170,11 +170,13 @@ func (mgr *locationManager) GetLocations() ([]LocationConfiguration, error) {
 }
 
 // GetLocationProperties returns properties configured for a given location
-func (mgr *locationManager) GetLocationProperties(locationName, infrastructureType string) (config.DynamicMap, error) {
+func (mgr *locationManager) GetLocationProperties(locationName, locationType string) (config.DynamicMap, error) {
 
-	if infrastructureType == adapter.AdaptedLocationType {
+	if locationType == adapter.AdaptedLocationType {
 		return mgr.hpAdapter.GetLocationConfiguration(locationName)
 	}
+
+	// get location configuration from the kv store
 	var props config.DynamicMap
 	kvp, _, err := mgr.cc.KV().Get(path.Join(consulutil.LocationsPrefix, locationName), nil)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


locations/location_mgr.go
```diff
@@ -170,9 +170,9 @@ func (mgr *locationManager) GetLocations() ([]LocationConfiguration, error) {
 }
 
 // GetLocationProperties returns properties configured for a given location
-func (mgr *locationManager) GetLocationProperties(locationName, locationType string) (config.DynamicMap, error) {
```
stebenoist: I think it's quite confusing to rename locationType into infrastructureType ?

adidanes: I looked to GetLocationPropertiesForNode usage and noted that its always called using an infrastructureType as the 4th parameter : GetLocationPropertiesForNode(ctx, deploymentID, nodeName, infrastructureType).
For example:
https://github.com/ystia/yorc/blob/bugfix/GH-550-check-loc-type/prov/terraform/google/generator.go#L55

adidanes: Terraform generators in particular use the infrastructure type found in the compute node.
It really correspond to a supported infrastructure type.

adidanes: The difficulty here is to cope with 2 _locationTypes_ : the argument of the function call and the type of the location identified by the locationName, also argument of the function call.
I made the choise to use 'locationType' for the second

stebenoist: Yep, externally for client, it 's an infrastructure type. But internally to the locations API, it's what has been called location type

adidanes: In this case, maybe we should change infrastructureType to locationType in the functions calling GetLocationPropertiesForNode in order to be coherent ?

stebenoist: Not sure it's really useful. WYT @loicalbertin @laurentganne  ?

laurentganne: actually we re-used the terms used by A4C which introduced the notion of location.
In A4C, a location is defined by  a name and an infrastructure type:
newLocation

stebenoist: Ok

stebenoist: I'm afraid we have now both infrastructureType and locationType in the locations package. This can be confusing

Issue #625, opened 2020-03-30
Location is not listed if its properties are missing


adidanes:
# Bug Report

## Description
When adding a new location configuration with the CLI command (yorc loc add), and its properties are empty, the location is created but it does not appears in the locations list printed by the CLI command (yorc loc list).

## Expected behavior
CLI command yorc loc list should print out the miss-configured location. 

## Steps to reproduce the issue

1. Step1
Add new location:

`yorc loc add -d '{"name": "myslurm", "type": "slurm", "properties": {}}'``

=> Location  myslurm  created

2. Step2 

`yorc loc list`

=> all other locations except 'myslurm'
## Output of `yorc version`
4.0.0-SNAPSHOT

## Priority
 Low

PR #626, opened 2020-03-30
List location name and type even when no properties provided

adidanes:
# Pull Request description

## Description of the change
Detect that a location properties are not provided, and add it to the outpul table.

### What I did
Add only location name and type to the output table.

### How to verify it
Add a location with no preperties:
`yorc loc add -d '{"name": "g1", "type": "google", "properties": {}}'`

Then list locations and check g1 is in the output table

## Applicable Issues
https://github.com/ystia/yorc/issues/625

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## UNRELEASED
 
+### BUG FIXES
+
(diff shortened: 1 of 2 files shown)


stebenoist:
Duplicate of #627

PR #643, opened 2020-05-07
Fix error on canceling a running deployment

### LINKED ISSUE #630: Unable to undeploy a deployment in progress from Alien4Cloud
# Bug Report

## Description
Unable to undeploy a deployment in progress from Alien4Cloud

## Expected behavior
During the deployment of an application, user could need to stop progress and undeploy.

## Actual behavior
In Alien4Cloud, after starting a Deploy operation, during status _Deployment in prrogress_, if user launches Undeploy, the application status pass in Failed, A4C prints _Undeployment in progress_ for ever, application is stucked.
The application can be deployed only by Yorc CLI command. 
I noticed with a multi node application (elk-basic) that compute instances on GCP may rest (not deleted by undeploy) if -p flag (purge) is not used.

Issue happens when deploying to GCP
Deploy / Undeploy of the test applicaion (elk-basic) on OpenStack succeded, but one of the compute instances was not deleted by the Undeploy (_Compute_KBN_ node ; it seems to me that this node instance has not completed createing when the workflow was cancelled: 
....Compute_KBN-0-ConnectionCheck: Still creating... (10s elapsed)
...

## Priority

Low

loicalbertin:
# Pull Request description

## Description of the change

Do not change task status when we just need to set the error message for task.
task status change should be handled in a unique place.

### What I did

Do not set task status on run step error.
Added a new function that allow to prevent overwriting of error message if a previous one already exist.

### How to verify it

Follow reproduction steps on #630 

* Unable to undeploy a deployment in progress from Alien4Cloud ([GH-630](https://github.com/ystia/yorc/issues/630))

## Applicable Issues

Fixes  #630 

Backport needed on release/4.0 (See #644)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 
 * Bootstrap may failed with a nil pointer error if download of a component fails ([GH-634](https://github.com/ystia/yorc/issues/634))
 * Missing concurrency limit during data migration for logs and events file storage ([GH-640](https://github.com/ystia/yorc/issues/640))
+* Unable to undeploy a deployment in progress from Alien4Cloud ([GH-630](https://github.com/ystia/yorc/issues/630))
 
 ## 4.0.0 (April 17, 2020)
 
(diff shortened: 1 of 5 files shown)


loicalbertin:
Yeah you are right, but this issue is not related to this bug.
I think this has been here for ages... And seems related to the way we store or not the Terraform state in Consul (to be checked)... We should open a different issue to track it.

PR #682, opened 2020-09-13
Migrate to new sortorder import path

fvbommel:
# Pull Request description

Due to some issues with the server handling the vanity imports, I unfortunately had to move the `vbom.ml/util` package to `github.com/fvbommel/util`. Since I was breaking import paths anyway, I decided to also move the much more popular `sortorder` sub-package to `github.com/fvbommel/sortorder` instead.

See fvbommel/util#7

## Description of the change

I've updated the import path to the new location.

### What I did

I've updated the import path where it was used.

### How I did it

I've updated the import path where it was used.

### How to verify it

The actual code of the imported package hasn't changed, [as can be seen in the Blame view](https://github.com/fvbommel/sortorder/blame/master/natsort.go).
The only changes to the .go files have been to remove the vanity import and to remove a test dependency.

Updated import path of vbom.ml/util/sortorder to github.com/fvbommel/sortorder

## Applicable Issues

None.

### DIFF
--- a/deployments/definition_store_test.go
+++ b/deployments/definition_store_test.go
@@ -29,8 +29,8 @@ import (
 	"strings"
 	"testing"
 	"time"
-	"vbom.ml/util/sortorder"
 
+	"github.com/fvbommel/sortorder"
 	ctu "github.com/hashicorp/consul/testutil"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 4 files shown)


loicalbertin (approved):
LGTM :+1: :clap:

Thank you very much @fvbommel, that's good to see someone that take care of downstream projects when renaming import paths! :+1:

fvbommel:
Yes, I felt kind of bad about that rename but I think it's the best solution in the long term. Helping out importers is my way of helping ease the pain :wink:.

PR #685, opened 2020-09-29
[Backport 4.0] Cleanup of ansible execution directories after execution

### LINKED ISSUE #683: Empty directories not removed after ansible executions can lead to inodes exhaustion
# Bug Report

## Description

When an operation is executed, Yorc creates this directory tree: `<working dir>/<deployment ID>/ansible/<task ID>_<timestamp>/<node name>`
and will create under this directory, subdirectories as well as files needed to run ansible playbooks.

Once the execution is done, a cleanup is done but not complete: this directory is not deleted, as well as a subdirectoty
So one single execution will consume several linux inodes, one for each of the remaining directories,
until the application is undeployed.

This can lead to inodes exhaustion, for example when the execution corresponds to the "run" operation of a job. This operation is executed every n seconds for the duration of the job. So for a job duration of several days, it represents a huge number of executions, and so a huge number of remaining directories (and consumed inodes).

## Priority

Medium

laurentganne:
# Pull Request description

### What I did

`prov/ansible/execution.go`
In the deferred function doing the cleanup after the playbook execution, deleting recursively the timestamped directory that was created for the execution

misc change not rleated to the issue: fixed two cases where a returned error was ignored.

`prov/ansible/execution_test.go`
Added a test checking directories created for an execution are deleted after the execution

### How to verify it

Deploy a topology with a compute instance and a component implementing standard and/or runnable interfaces.
Check during the workflow execution that as soon as a workflow step executing a component operation is executed, the corresponding timestamped directories under `<working dir>/<deployment ID>/ansible/`is removed.
Check once the workflow is done, that there is no timestamped directory under `<working dir>/<deployment ID>/ansible/`.

Empty directories not removed after ansible executions can lead to inodes exhaustion ([GH-683](https://github.com/ystia/yorc/issues/683))

## Applicable Issues

* Fixes #683
* Backported from #684

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## UNRELEASED
 
+### BUG FIXES
+
+* Empty directories not removed after ansible executions can lead to inodes exhaustion ([GH-683](https://github.com/ystia/yorc/issues/683))
+
 ## 4.0.3 (August 27, 2020)
 
 ### FEATURES
--- a/prov/ansible/execution.go
+++ b/prov/ansible/execution.go
@@ -895,15 +895,16 @@ func (e *executionCommon) executeWithCurrentInstance(ctx context.Context, retry 
 		return err
 	}
 
-	ansibleRecipePath := filepath.Join(ansiblePath, stringutil.UniqueTimestampedName(e.taskID+"_", ""), e.NodeName)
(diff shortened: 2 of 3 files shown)


loicalbertin (approved):
LGTM :+1: :clap:

PR #696, opened 2020-10-13
uninstall Ansible older version that cannot be upgraded

### LINKED ISSUE #695: Bootstrap fails on hosts where a version of ansible < 2.10.0 is installed
# Bug Report

## Description

Continuous integrations tests for Yorc 4.1.x fail when performed on a Hosts Pool, because the same Hosts Pool is used for Yorc 4.0.x and this version is installing an earlier version of ansible (2.7.9), while 4.1.0 is installing ansible 2.10.0.
The upgrade to ansible 2.10.0 from an earlier version is not supported and fails, and so the bootstrap of Yorc 4.1 on a Hosts Pool fails when it attempts to upgrade the Ansible 2.7.9 installed by Yorc 4.0.x.
The Yorc 4.1 bootstrap should take care of uninstalling ansible with a version < 2.10.0 before attempting to install Ansible 2.10.0

## Priority

Medium 

laurentganne:
# Pull Request description

### What I did

Updated the playbook installing ansible, for the case where ansible has to be upgraded from a version prior to 2.10.0, and for the case when it needs to be upgraded again to 2.10.0 after a downgrade.
See corresponding changes in Forge AnsibleRuntime component done in pull request https://github.com/ystia/forge/pull/139

### How to verify it

Run continuous integration tests on this pull request on a Hosts Pool, this will install ansible 2.10.0
Run continuous integration tests on release/4.0 branch, this will downgrade ansible to 2.7.9
Run continuous integration tests on this pull request on a Hosts Pool, this will upgrade ansible to 2.10.0 by uninstalling ansible 2.7.9, uninstalling ansible-base to avoid an issue with a missing ansible-playbook binary, and installing ansible 2.10.0

Bootstrap fails on hosts where a version of ansible < 2.10.0 is installed ([GH-695](https://github.com/ystia/yorc/issues/695))

## Applicable Issues

Closes #695


loicalbertin (approved):
LGTM :+1: :clap:

PR #699, opened 2020-10-15
Stop printing ES config at startup (so verbose when TLS enabled)

### LINKED ISSUE #681: Remove useless/cluttering logs
# Issue Type

* [ ] Bug report
* [X] Improvement request

## Description
Since we have deployment logs stored in ES,(yorc 4.0.2),  we have a lot of logs that are not useful at all and are just using space for no valid reason (from our point of view)

This log : 
2020/09/10 15:49:09 [INFO]  #011- Will use this ES client configuration: 
is OK, but inside there is the CACert:[ which outputs a *very* long sequence of integers, which make a very long log line at every restart that is not useful.

Also, during run we get those lines : 
 List called result k: _yorc/events/, waitIndex: 1599745604960997966, timeout: 5m0s, LastIndex: 1599745612445439556, len(values): 32
2020/09/10 15:46:58 [INFO]  List called result k: _yorc/logs/, waitIndex: 1599745605074028380, timeout: 5m0s, LastIndex: 1599745612647486273, len(values): 59
2020/09/10 15:46:58 [INFO]  List called k: _yorc/events/, waitIndex: 1599745612445439556, timeout: 5m0s
2020/09/10 15:46:58 [INFO]  List called k: _yorc/logs/, waitIndex: 1599745612647486273, timeout: 5m0s
They seem pretty useless too. (At least if you don't remove them, please put them in DEBUG level) 

1. Step1
2. Step2 ...

## Priority

Low

xdegenne:
# Pull Request description

ES Store was printing it's config at startup, generating too much output almost when TLS is activated.

### What I did

We'll now print ES config only in debug mode

Fixes #681

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@
 
 ### ENHANCEMENTS
 
+* Remove useless/cluttering logs ([GH-681](https://github.com/ystia/yorc/issues/681))
 * Should be able to specify edcsa or rsa ssh keys in gcloud compute instances metadata ([GH-697](https://github.com/ystia/yorc/issues/697))
 * Add the ability to define OpenStack Compute Instance metadata ([GH-687](https://github.com/ystia/yorc/issues/687))
 * Support Alien4Cloud 3.0 ([GH-689](https://github.com/ystia/yorc/issues/689))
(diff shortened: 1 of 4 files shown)


laurentganne (approved):
Looks good!

PR #701, opened 2020-10-16
[Back-port 4.0] Support ssh connection retries

### LINKED ISSUE #688: Support ssh connection retries
For deployments on Slurm or Hosts Pools, a single ssh connection failure due to an intermittent issue on the network can cause a workflow failure.
This can have bad consequences, like having a workflow with a Slurm long running job fail after several days of run just because Yorc faced an intermittent ssh connection issue, while the job is still running without any issue on Slurm.

This could be implemented in Yorc sshutil module at https://github.com/ystia/yorc/blob/v4.0.3/helper/sshutil/sshutil.go#L139
used by both Slurm and Hosts Pools.

loicalbertin:
# Pull Request description

### What I did

Implemented a retry mechanism for ssh connections

Also slightly changed the connection timeout to just target time for opening a connection not whole time to execute the command

Added the possibility to overwrite general ssh connection timeout and retries in Slurm location config.

* Support ssh connection retries ([GH-688](https://github.com/ystia/yorc/issues/688))

## Applicable Issues

Closes #688 
Back-ported to branch release/4.0 from #700


laurentganne (approved):
ok!

PR #711, opened 2021-01-06
Implement a synchronous purge endpoint

### LINKED ISSUE #707: Add a new synchronous purge API endpoint
This new purge endpoint should be synchronous (blocking no task)
There is 2 options for Purge:

* classic/default : Yorc checks that the deployment is in UNDEPLOYED state before starting a purge. If purge encounter an error then the deployment is placed in PURGE_ERROR state
* forced purge: Yorc doesn't check the deployment status or even if the deployment exists, Yorc removes everything it can

The purge process does not fast-fail: if we encounter an error,  we keep going but we finally  generate a purge report on what went wrong. (in force mode for sure, to be checked in classic mode)
* If there is at least one error, it should be to the user, with ability to  rerun the purge
* Error report available in Yorc system logs

loicalbertin:
# Pull Request description

### What I did

- Added a new API endpoint `POST /deployments/<id>/purge[?force]`
- Added a purge command on CLI
- Added a PURGE_FAILED deployment status

### How to verify it

1. Deploy an application
2. ensure that alien is not running as it may auto-purge the application using the old way
3. use the CLI to try to purge the deployed application => you should not be able to do it as deployment is not in `undeployed` state
4. undeploy the deployment using the CLI without the purge option
5. change the owner of the `work/deployments/<depid>` folder in order to generate an error when purging the deployment
6. use the CLI to try to purge the deployed application => Yorc should failed
7. use the CLI to **force** purge the deployed application => deployment should be purged (do not appear anymore in `yorc d list`) but with an error
8. change back the owner in order to allow Yorc to remove the  `work/deployments/<depid>` folder
9. use the CLI to **force** purge the deployed application => purge should be OK and the  `work/deployments/<depid>` folder should not exist anymore

* Add a new synchronous purge API endpoint ([GH-707](https://github.com/ystia/yorc/issues/707))

## Applicable Issues

Fixes #707


rest/deployments.go
```diff
@@ -397,3 +399,32 @@ func (s *Server) listDeploymentsHandler(w http.ResponseWriter, r *http.Request)
...
+				errs.Errors = append(errs.Errors, &Error{"internal_server_error", http.StatusInternalServerError, "Internal Server Error", err.Error()})
+			}
+		} else {
+			errs.Errors = append(errs.Errors, newInternalServerError(err))
+		}
+
+	}
+	if len(errs.Errors) > 0 {
+		w.WriteHeader(http.StatusInternalServerError)
+	}
```
laurentganne, line 428: set the status in case of success ? 
} else {
    w.WriteHeader(http.StatusOK)
}

laurentganne (approved):
Looks good!

rest/deployments.go
loicalbertin: This is not necessary: 
> 	// If WriteHeader has not yet been called, Write calls
>	// WriteHeader(http.StatusOK) before writing the data. 
<https://pkg.go.dev/net/http#ResponseWriter>

loicalbertin: But you are right if we do not set it we do not record it in the telemetry handler

PR #712, opened 2021-01-22
Fix consul path when creating a google subnet

trihoangvo:
# Pull Request description

When the topology has a node from type `yorc.nodes.google.Subnetwork`, yorc throws an error when storing runtime attributes of google subnet, because the consul path `/topology/nodes/<node_name>` does not follow the regex defined in:
https://github.com/ystia/yorc/blob/9d46bb2982f7598fa55f2a66f91d8dad08f21d60/deployments/attribute_notifications.go#L78

### What I did

Update consul path from `/topology/nodes/<node_name>` to `/topology/instances/<node_name>/0`

### How to verify it

Create a VPC network and set the property `custom_subnetworks`.


loicalbertin:
Hi @trihoangvo,

Thanks for this bug fix. Could you please rebase your PR on https://github.com/ystia/yorc/tree/upstream/fix_consul_path_googoe_subnet?
I added few refactorings.

loicalbertin:
Closed in favor for #714

PR #721, opened 2021-04-02
Some ES Store improvements

### LINKED ISSUE #722: Alllow shards and replicas configuration for Elastic storage
Should implement the following enhancements to Elastic storage:

* Allow shards and replicas configuration at index creation
* Fix a bug where the Store is still querying ES after the context has been cancelled
* Some logging improvements when ES is returning error messages

SebastienLaurent:
## Description of the change

Some general improvments on ES Store

### What I did
- Allowing shards and replicas configuration at index creation
- Fix a bug where the Store is still querying ES after the context has been cancelled
- Some logging improvements when ES is returning error messages

### How to verify it
The new log messages can be tested with a bad condition ES cluster  (red indices, insufficient nodes started,...)

Fixes #722


laurentganne (approved):
Looks good!
Would just need to add this line in `./CHANGELOG.md`section `ENHANCEMENTS`:

* Alllow shards and replicas configuration for Elastic storage ([GH-722](https://github.com/ystia/yorc/issues/722))

PR #734, opened 2021-05-06
Bugfix/gh 733 action no stop

### LINKED ISSUE #733: Workflow with asynchronous action never stops after another step failure
# Bug Report

## Description

A Workflow is running at some point two steps in parallel:
* a standard lifecycle "create" operation on a component
* a job lifecyle asynchronous operation "run" on a job.

The "create" operation fails. Then the corresponding step appears as failed in the workflow, as expected.
But the "run" step of the job component will always appear as running, and no new action to run again the run action is scheduled.
The workflow is seen as running, forever.

To workaround this issue, the administrator will have to change the status of the yorc task manually.
Before this manual change, the internal statuses for tasks and workflow in consul show that:
* the task status is 1 (RUNNING), 
* the workflow step corresponding to the create operation of the standard lifecycle component is ERROR
* the workflow step corresponding to the run action of the job component is RUNNING.
* there is no scheduling action registered (as it has been unregistered by yorc because of the other step that failed in the workflow), but there still one single entry `_yorc/scheduling/actions/<id>/data/<key>` in consul corresponding to an action data that the run operation has updated while the corresponding action had already been unregistered by yorc.

## Priority

Medium

laurentganne:
# Pull Request description

## Description of the change

Fixed an issue where a workflow will be seen running forever when a synchronous step failed while an asynchronous run action is taking place at the same time. The asynchronous action was unscheduled as expected but its status was not updated and was seen as running forever.

Fixed also an issue when the asynchronous action is updating its action data: there was no check to see if the action has been unregistered, and the action data was added, leaving an isolated key in consul.

### What I did

`prov/scheduling/scheduler/consul_test.go`
`prov/scheduling/scheduler/scheduler_test.go`:
Added a test on UpdateActionData()

`prov/scheduling/scheduling.go`:
Updating action data only when the action data deploymentID key exists

`tasks/workflow/worker.go`:
On task failure, call endAction()

Workflow with asynchronous action never stops after another step failure  ([GH-733](https://github.com/ystia/yorc/issues/733))

## Applicable Issues

Closes #733

### DIFF
--- a/prov/scheduling/scheduling.go
+++ b/prov/scheduling/scheduling.go
@@ -22,7 +22,7 @@ import (
 
 	"github.com/hashicorp/consul/api"
 	"github.com/pkg/errors"
-	"github.com/satori/go.uuid"
+	uuid "github.com/satori/go.uuid"
 
 	"github.com/ystia/yorc/v4/helper/consulutil"
 	"github.com/ystia/yorc/v4/log"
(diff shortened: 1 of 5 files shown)


prov/scheduling/scheduling.go
```diff
@@ -104,7 +104,16 @@ func UnregisterAction(client *api.Client, id string) error {
 // UpdateActionData updates the value of a given data within an action
 func UpdateActionData(client *api.Client, id, key, value string) error {
 
-	//TODO check if action exists
-	scaKeyPath := path.Join(consulutil.SchedulingKVPrefix, "actions", id, "data", key)
+	// check if action still exists
```
loicalbertin, line 107: Nice to fix TODOs :+1:  :wink:

loicalbertin (approved):
LGTM :+1: :clap:

PR #746, opened 2021-06-11
Fix Consul connections over consumtion on task monitoring

### LINKED ISSUE #745: Over-consumption of Consul connections
# Bug Report

## Description

When Yorc runs a large number of async operation (typically for jobs monitoring) we noticed an over-consumption of connections to Consul. This is due to the way we monitor tasks errors and cancellation.
We use a long poll connection on tasks flags but we do not properly use context cancellation to detect that we could stop the monitoring. 
https://github.com/ystia/yorc/blob/e8759d1766446d6246bf20f8e5a0a0dcbf40f240/tasks/tasks.go#L636
This leads to consul connections being still waiting for an HTTP timeout to be released while they are not useful anymore.
When action scheduling interval is low or in case of many different actions then we can consume a very large number of Consul connections on this.

loicalbertin:
# Pull Request description

### What I did

Just added the context to the consul query options.
This context is then passed to the http request used by the Consul client and finally can be used for request cancellation

### How to verify it

Generate a high load of job monitoring actions for exemple by using a monitoring interval of `0.5s` and monitor the consul connections using for instance `lsof`. It should remain constant

* Over-consumption of Consul connections ([GH-745](https://github.com/ystia/yorc/issues/745))

## Applicable Issues

Fixes #745
Backported to 4.1.x in PR #747
Backported to 4.1.x in PR #748

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 
 ### BUG FIXES
 
+* Over-consumption of Consul connections ([GH-745](https://github.com/ystia/yorc/issues/745))
 * Yorc panics attempting to print an error handling a script execution stdout ([GH-741](https://github.com/ystia/yorc/issues/741))
 * Error submitting a SLURM job with no execution option ([GH-739](https://github.com/ystia/yorc/issues/739))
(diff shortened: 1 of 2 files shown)


laurentganne (approved):
Perfect!

PR #770, opened 2021-07-26
bootvolume sourcetype was a requirement

tibeer:
Signed-off-by: Tim Beermann <beermann@osism.tech>

# Pull Request description

1. Filling out the template is required. Any pull request that does not include enough information to be reviewed in a timely manner may be closed at the maintainers' discretion.
2. Please read our contribution guidelines
3. PR with a decent set of unit tests have more chance to be reviewed and merged quickly

## Description of the change

Yorc's openstack BootVolume type required the property **source**. This prevented e.g. importing topologies which did not have a BootVolume set.
Leaving **source** as a requirement would force setting a boot volume in terraform resulting in automatically requiring an *image_id* if you want
to boot from an image. Thogh if you want to boot a VM from Terraform from an image name you must leave out the boot volume OR query the *image_id*
as a data source upfront. Anyway, the **source** attribute should never be a requirement.

### What I did

Changed the required value from `true` to `false` for *yorc.datatypes.openstack.BootVolume*'s property **source**.

### How I did it

none

### How to verify it

none

OpenStack type *yorc.datatypes.openstack.BootVolume* required the source propery. This prevented creating VMs without one directly from image names.

## Applicable Issues

none

### DIFF
--- a/data/tosca/yorc-openstack-types.yml
+++ b/data/tosca/yorc-openstack-types.yml
@@ -18,7 +18,7 @@ data_types:
         description: UUID of the image, volume, or snapshot. Required unless source is set to 'blank'
       source:
         type: string
-        required: true
+        required: false
         description: Source type of the device. Must be one of 'blank', 'image', 'volume', or 'snapshot'
         constraints:
           - valid_values: [ blank, image, volume, snapshot ]


laurentganne:
This issue should have been introduced by a change in Alien4Cloud management of required properties having default values.
With previous versions of Alien4Cloud, you could either define the property `ìmageName` or `image` and not define at all the property `boot_volume` as this property `boot_volume` is declared as not required.
And this worked well.

Now, Alien4cloud will systematically expect the `boot_volume` property to be defined the topology although it is declared as not required, apparently because this property `boot_volume` has a property `delete_on_termination` which is declared as required and has a default value.

I just saw this as well on my setup during the substitution of abstract resource with on-demand resources, where Alien4Cloud is systematically adding a property `boot_volume` in the topology with the default value of the required property `delete_on_termination`.
Then Alien4Cloud will complain during the deployment that the required property source has no value.

An alternate solution, will need to check if it works as expected:
I think to deal with this new behavior of Alien4Cloud regarding required properties with default values, it would be better to keep `source` as a required value, which is conform to the terraform specification of a block device.
And make `delete_on_termination` as a non-required property, so that we would have this expected behavior:
whenever the user specifies an `ìmageName`, Alien4Cloud should not expect that a property `boot_volume` is defined when you import a topology, and Alien4Cloud should not add such boot_volume property to your topology as well when it is substituting abstract resources with on-demand resources during a deployment.

tibeer:
Older than one year. Closing this PR.

Issue #771, opened 2021-09-10
Allow to replay workflow steps even if they are not in error


laurentganne:
Following pull request #754 allowing to replay workflow steps in error, it is also interesting to allow to replay workflow steps that are not in error.

For example, in the lifecycle operations of a job, this job is created, submitted, then a step "run" performs the asynchronous monitoring of the job status, until the job ends on success or on failure.
If the job fails, the run "step" will fail.
But there is no interest in replaying this "run" step, that is just monitoring a job status.

To replay the workflow, the user would need to be able to resubmit the job, although the step submitting the job was successful in the first execution.

laurentganne:
Merge #772

PR #772, opened 2021-09-10
Feature/job replay

laurentganne:
# Pull Request description

Added the ability to replay workflow steps, even if they were successful. This can be used to replay the successful submission of a job after a failure during the execution of this job.
 
### What I did

`rest/http_api.md`
`tasks/tasks.go`
allowing to change a DONE step (to resubmit a job that failed during its execution for example)

`tasks/workflow/step.go`
Not ending the workflow on a step failure if this step has "on failure" next steps

`tasks/workflow/worker.go`
endAction() and runAction(): should not notify the task error if there are on failure steps, to go on in the workflow executing these on failure steps
runWorkflowStep(): on workflow replay, when an asynchronous step is done, its next steps must be registered

`tasks/workflow/consul_test.go`
`tasks/workflow/testdata/workflow.yaml`
`tasks/workflow/worker_test.go`

Added a test checking replaying a workflow containing workflow with an asynchronous step already done, will now execute steps following this asynchronous step

### How to verify it

Tested on a LEXIS step with a job performing data transfers between Compute locations:
Using a workflow containing job steps, make the job executing a data transfer fail.
Then use yorc command :
```bash
yorc deployments tasks update-step-state <deployment ID> <task ID> <step name> initial
```
to re-iniitlalise the run step that was on error and the submit step that was successful.

Then use yorc command:
```bash
yorc deployments tasks resume <deployment ID> <task ID>
```
to replay the workflow.
Check the job is resubmitted, then monitored until its end.

Allow to replay workflow steps even if they are not in error ([GH-771](https://github.com/ystia/yorc/issues/771))

## Applicable Issues

#771

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,8 @@
 
 ### FEATURES
 
-* Workflows steps replays on error  ([GH-753](https://github.com/ystia/yorc/issues/753))
+* Allow to replay workflow steps even if they are not in error ([GH-771](https://github.com/ystia/yorc/issues/771))
+* Workflows steps replays on error ([GH-753](https://github.com/ystia/yorc/issues/753))
 
 ### ENHANCEMENTS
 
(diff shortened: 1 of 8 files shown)


loicalbertin (approved):
LGTM :+1: :clap:

PR #776, opened 2022-01-12
Ability to authenticate against OpenStack with application credentials or token

### LINKED ISSUE #775: Add ability to authenticate to OpenStack with token or application credentials provided by third party
In some environments, like in LEXIS project, Yorc must be able to allocate OpenStack compute resources on behalf of any user created on demand by a third party AAI (Authentication and Authorization Infrastructure), Yorc being just given an OpenStack token or Openstack application credentials, valid only for a given time defined by the third party AAI.

It should be possible to only rely on an OpenStack token, even in the case of application credentials as it is possible to generate an OpenStack token from OpenStack application credentials,
but there is currently a bug in OpenStack identity (keystone) tracked by https://bugs.launchpad.net/keystone/+bug/1878438 "error when using token get from application credential authentication".
When attempting to use such a token, this failure occurs in keystone:

`ERROR keystone   File "/openstack/venvs/keystone-21.2.2.dev1/lib/python3.6/site-packages/keystone/api/_shared/authentication.py", line 212, in authenticate_for_token`

So, in addition to tokens for a future use, we should be able to directly use application credentials as well until https://bugs.launchpad.net/keystone/+bug/1878438 will be addressed.
The corresponding values having an ephemeral validity managed by a third party software, per user, are not to be stored in Yorc configuration.
They can be provided in the OpenStack node template metadata, and read/used by Yorc when it attempts to allocate/release the corresponding OpenStack resource.

laurentganne:
# Pull Request description

## Description of the change

In some environments, like in LEXIS project, Yorc must be able to allocate OpenStack compute resources on behalf of any user created on demand by a third party AAI (Authentication and Authorization Infrastructure), Yorc being just given an OpenStack token or Openstack application credentials, valid only for a given time defined by the third party AAI.

Added the ability for Yorc to authenticate against OpenStack using a token or application credentials, that are provided in a node template metadata (not in yorc static configuration as these are per user values with a validity defined by an external AAI. 

### What I did

`doc/configuration.rst`

In the table of OpenStack parameters,
added parameter domain_id to be used with tokens, and referencing a note for user_name and password.
The note added below the table specifies that a token or application credentials can be specified in a node template, when user/password authentication can't be used.

`tosca/constants.go`
Added constants for token and application credentials keys that can be define in a node template metadata

`prov/terraform/openstack/generator.go`

When setting OpenStack parameters used for authentication, checking if application credentials are defined in the node template metadata to use them during the authentication,
else do a token-based authentication if a toekn is defined, else do the user/password authentication like previously.

### How to verify it

Verified on a LEXIS setup, using both:
* the usual user/password static configuration
* the external AAI provide application credentials, that a Yorc plugin (https://github.com/lexis-project/yorc-dynamic-orchestration-plugin) injects in  node template metadata.

Ability to authenticate against OpenStack with token or application credentials ([GH-775](https://github.com/ystia/yorc/issues/775))

## Applicable Issues

Closes #775

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### FEATURES
 
+* Ability to authenticate against OpenStack with token or application credentials ([GH-775](https://github.com/ystia/yorc/issues/775))
 * Allow to replay workflow steps even if they are not in error ([GH-771](https://github.com/ystia/yorc/issues/771))
 * Workflows steps replays on error ([GH-753](https://github.com/ystia/yorc/issues/753))
 
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -1010,13 +1010,15 @@ OpenStack location type is ``openstack`` in lower case.
 +-----------------------------------+---------------------------------------------------------------------------------------------------------------------+-----------+----------------------------------------------------+---------------+
 | ``user_domain_name``              | Specify the domain name where the user is located (Identity v3 only).                                               | string    | yes (if use Identity v3)                           |               |
 +-----------------------------------+---------------------------------------------------------------------------------------------------------------------+-----------+----------------------------------------------------+---------------+
+| ``domain_id``                     | Specify the ID of the Domain to scope to (Identity v3 only).                                                        | string    | no                                                 |               |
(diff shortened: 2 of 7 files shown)


loicalbertin (approved):
LGTM :+1: :clap:

PR #782, opened 2022-02-07
Fixed issue when using existing BlockStorage for OpenStack by volume_id

trihoangvo:
# Pull Request description

When we set `volume_id` in BlockStorage OpenStack, yorc reads the property but does not use it.

### What I did

In the implementation, the `volumeID` variable is declared but never assigned from `volumeIDValue` variable so we assign it correctly.

### How to verify it

Deploy a topology in OpenStack with an existing block storage set with `volume_id`.

Fixed issue when using existing BlockStorage for OpenStack by volume_id

## Applicable Issues

no


laurentganne (approved):
Looks good!

loicalbertin (approved):
LGTM :+1: :clap:

PR #790, opened 2022-11-18
Fix `get_property`/`get_attribute` functions with `REQ_TARGET` keyword failing even if requirement is optional

### LINKED ISSUE #789: `get_property`/`get_attribute` functions should not fail with `REQ_TARGET` keyword if requirement is optional and not actually linked on a node template with a relationship template
# Bug Report

## Description

`get_attribute: [ REQ_TARGET, requirement_name, attribute_name]` and `get_property: [ REQ_TARGET, requirement_name, property_name]` fails to resolve if an optional requirement is not actually connected to a capability.

## Expected behavior

If requirement definition is defined as optional and there is no actual relationship template on the requirement then `get_property`/`get_attribute` functions should not fail and return no value

loicalbertin:
# Pull Request description

## Description of the change

Added a check that the requirement is not optional before failing.
Note that this is very unlikely to happen when using Alien4Cloud as it enforces this rule.

### What I did

If target node is not found and requirement is optional then return an empty value without error

*  Fix `get_property`/`get_attribute` functions with `REQ_TARGET` keyword failing even if requirement is optional  ([GH-789](https://github.com/ystia/yorc/issues/789))

## Applicable Issues

Fixes #789

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -83,7 +83,7 @@ jobs:
 
       - name: Install sphinx dependencies
         run: |
-          pip install --user --upgrade sphinx==1.8.1 semantic-version requests urllib3[secure]==1.23 Pygments>=2.7.1
+          pip install --user --upgrade sphinx<2 semantic-version requests urllib3[secure]==1.23 Pygments>=2.7.1
           pip install -r doc/requirements.txt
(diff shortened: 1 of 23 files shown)


laurentganne (approved):
Looks good!

PR #792, opened 2022-11-22
Support nested TOSCA functions in `get_input`

### LINKED ISSUE #791: Support nested TOSCA functions in get_input
**As:** TOSCA Component dev
**I want to:** use the `get_input` TOSCA function with nested TOSCA functions 
**So that:** I can build complex stuff like `{ get_input: [ concat: [get_property: [SELF, input_base_name], "_v1"]]}`

**AC1:** Similarly to #777,  a limitation in Alien4Cloud prevent to use something else than a scalar as parameter of the `get_input` function.  Apart from that this feature is already functional in Yorc. So we propose to introduce an alias to the `get_input` function called `get_input_nf` (`nf` for `nested functions`) that will not be handled and rejected by Alien.

loicalbertin:
# Pull Request description

## Description of the change

The issue was mainly that Alien4Cloud does not support nested functions in `get_input` while Yorc itself does.

### What I did

Created an alias for `get_input` called `get_input_nf` allowing to use this function in TOSCA types that are designed to be used in Alien4Cloud

* Support nested TOSCA functions in get_input (added a get_input_nf function) ([GH-791](https://github.com/ystia/yorc/issues/791))

## Applicable Issues

Fixes #791

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 
 * Ability to authenticate against OpenStack with token or application credentials ([GH-775](https://github.com/ystia/yorc/issues/775))
 * Support nested TOSCA functions in get_secret (added a get_vault_secret function) ([GH-777](https://github.com/ystia/yorc/issues/777))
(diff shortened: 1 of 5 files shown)


laurentganne (approved):
Looks good!

